PR #15, opened 2018-12-11
Golfing

lukeed:
Sorry for the long-ish PR, but I wanted to break apart each step to document its savings as well make it easier to discuss/revert individual changes. If this is to be accepted, I'd recommend squashing.

> Also, the spacing is a mess 🙈 The `src/.editorconfig` looks to be misconfigured or in the wrong place & I didn't want to adjust it as it'd add PR noise, making the 🏌️hard to see.

The core functionality is preserved at **735 B (gz) / 601 B (br)** – down from **902 B (gz) / 747 B (br)**. It's possible to get it down to `717 gz / 594 br` by inlining the `options` defaults, but that may not read as well 😄 

There _is_ 1 breaking change (which, of course, I can revert): The `priority` option is now a Boolean. Not only did this save ~30 bytes, but IMO it makes more sense since the only options were "high" vs "low". In this format, `options.priority=true` is "high" priority.

The other big wins came from:

* turning the `AsyncFunction` into a regular function that returned the Promise
* dropping `link.setAttribute`s in favor of using `link[key]` directly
* using `document.head` and `document.querySelector`

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -42,14 +42,14 @@ function prefetcher(url) {
  * @param {Object} options - Configuration options for quicklink
  * @param {Array} options.urls - Array of URLs to prefetch (override)
  * @param {Object} options.el - DOM element to prefetch in-viewport links of
- * @param {string} options.priority - Attempt to fetch with higher priority (low or high)
+ * @param {Boolean} options.priority - Attempt higher priority fetch (low or high)
  * @param {Number} options.timeout - Timeout after which prefetching will occur
  * @param {function} options.timeoutFn - Custom timeout function
  */
 export default function (options) {
   options = Object.assign({
-    priority: 'low',
     timeout: 2e3,
+    priority: false,
     timeoutFn: requestIdleCallback,
     el: document,
   }, options);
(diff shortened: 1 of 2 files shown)


src/index.mjs
```diff
@@ -42,14 +42,14 @@ function prefetcher(url) {
  * @param {Object} options - Configuration options for quicklink
  * @param {Array} options.urls - Array of URLs to prefetch (override)
  * @param {Object} options.el - DOM element to prefetch in-viewport links of
- * @param {string} options.priority - Attempt to fetch with higher priority (low or high)
+ * @param {Boolean} options.priority - Attempt higher priority fetch (low or high)
```
addyosmani: I initially planned for priority to have a tri (or more) state set of values, but it's very likely to stay at either high or low. With that in mind, strongly in favor of switching to a `Boolean` here 👍

lukeed: Ah, gotcha. I didn't even know there _could_ be other priorities 😆

addyosmani:
> src/.editorconfig

I think this snuck in after adopting microbundle 😋. I'll have to fix up or ignore after landing this change.

lukeed:
Haha, it's happened to me before too. I can fix it up for you after if you'd like.

lukeed:
Awesome, thank you! 🙇

PR #17, opened 2018-12-11
fix: update link to Guess.js

bjarneo:
Not sure if the site is down or not, but created this PR in case the link is wrong.


bjarneo:
Closing as http://guessjs.com works, might just be issues with ssl.

PR #23, opened 2018-12-12
Implement limit

jadjoubran:
Implemented limit 😃

Things I'm not sure about:
* default limit to `Infinity`? on line 56 where options is being merged with the defaults
* create a new variable rather than modifying `entries`
* create a new demo & test file for `limit` use cases?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ The above options are best for multi-page sites. Single-page apps have a few opt
 * `priority`: Boolean specifying preferred priority for fetches. Defaults to `false`. `true` will attempt to use the `fetch()` API where supported (rather than rel=prefetch)
 * `origins`: Static array of URL hostname strings that are allowed to be prefetched. Defaults to the same domain origin, which prevents _any_ cross-origin requests.
 * `ignores`: A RegExp, Function, or Array that further determines if a URL should be prefetched. These execute _after_ origin matching.
+* `limit`: Limits the number of links to be prefetched. Defaults to `Infinity`.
 
 TODO:
 * Explore detecting file-extension of resources and using [rel=preload](https://w3c.github.io/preload/) for high priority fetches
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -20,6 +20,7 @@ import requestIdleCallback from './request-idle-callback.mjs';
 const toPrefetch = new Set();
 
 const observer = new IntersectionObserver(entries => {
+  entries = entries.slice(0, observer.limit);
   entries.forEach(entry => {
     if (entry.isIntersecting) {
       const url = entry.target.href;
@@ -60,6 +61,7 @@ export default function (options) {
   }, options);
 
   observer.priority = options.priority;
+  observer.limit = options.limit;
 
   options.timeoutFn(() => {
     // If URLs are given, prefetch them.
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -20,7 +20,6 @@ import requestIdleCallback from './request-idle-callback.mjs';
 const toPrefetch = new Set();
 
 const observer = new IntersectionObserver(entries => {
-  entries = entries.slice(0, observer.limit);
   entries.forEach(entry => {
     if (entry.isIntersecting) {
       const link = entry.target;
@@ -71,6 +70,7 @@ function isIgnored(node, filter) {
  * @param {Array|RegExp|Function} options.ignores - Custom filter(s) that run after origin checks
  * @param {Number} options.timeout - Timeout after which prefetching will occur
  * @param {Function} options.timeoutFn - Custom timeout function
+ * @param {Number} options.limit - Limit the number of links to be observed
  */
 export default function (options) {
   options = Object.assign({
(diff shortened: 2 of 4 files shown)


jadjoubran:
Yes sure, here's what I had in mind, assuming quicklink has been initialized with a `limit: 5`

You scroll through a page that happens to have a lot of links that are all visible in the viewport (say more than 20), quicklink will only prefetch 5 links out of these 20
If you scroll again and you encounter another 30 links, you will only prefetch 5

Whereas your approach is about observing the first 5 links of the entire page.

I don't have a strong opinion on my approach, so which one do you think would be more useful?

lukeed:
I would have thought that the limit applied to the `prefetch` itself – aka, set max 5 network requests at once

addyosmani:
I'm thinking through the user flow here: so, I load a page and we begin prefetching in-viewport links and limit to 5. If I scroll through the next batch of links are also limited to five. With limiting links to inspect, I think we risk an aggregation problem - you have 5, then 10, then 15 resources being prefetched if you scroll down a page. Is that correct?

Limiting dispatched network requests doesn't have this problem as at any time you can at most only prefetch 5 resources. I'm inclined to go for this approach personally.

jadjoubran:
Yes what you said is correct 👍 
okay I'll send a PR tomorrow morning, just to be clear, is the below correct?
Sorry but it's quite complicated as we're talking about a hypothetical scenario

* `limit` would limit the result returned by `options.el.querySelectorAll('a')` so that if the user limits to 5, only 5 links will be observed (and then potentially prefetched while scrolling)

Is that correct?
Thanks 😄

addyosmani:
Related: I just saw https://github.com/GoogleChromeLabs/quicklink/pull/56 come in :)

Yeah, we can go for limiting the number of items to be prefetched based on the queried results from qSA. It would also be great to have tests/docs for the change if possible.

jadjoubran:
Okay this is done from my side, looking for some feedback

I also have a couple of questions:
* should `limit` also be applied to options.urls?
* Does `Inifinity` make sense as a default?

src/index.mjs
```diff
@@ -71,17 +70,18 @@ function isIgnored(node, filter) {
...
  * @param {Function} options.timeoutFn - Custom timeout function
+ * @param {Number} options.limit - Limit the number of links to be observed
  */
 export default function (options) {
   options = Object.assign({
     timeout: 2e3,
     priority: false,
     timeoutFn: requestIdleCallback,
     el: document,
+    limit: Infinity
```
addyosmani: I'm OK with `Infinity` being the default for `limit`. A minified version of `Infinity` would be `1/0`, which I want to say `microbundle` should handle for us. cc @lukeed 

An alternative design could have been only accepting a `limit` value if it wasn't undefined, but would probably lead to a longer implem :)

README.md
```diff
@@ -82,6 +82,7 @@ The above options are best for multi-page sites. Single-page apps have a few opt
 * `priority`: Boolean specifying preferred priority for fetches. Defaults to `false`. `true` will attempt to use the `fetch()` API where supported (rather than rel=prefetch)
 * `origins`: Static array of URL hostname strings that are allowed to be prefetched. Defaults to the same domain origin, which prevents _any_ cross-origin requests.
 * `ignores`: A RegExp, Function, or Array that further determines if a URL should be prefetched. These execute _after_ origin matching.
+* `limit`: Limits the number of links to be prefetched. Defaults to `Infinity`.
```
addyosmani: I think we'll want to provide more detail on `limit`. Folks may be wondering if this is the total limit of links quicklink will ever prefetch in a session or if it's the number of links that will be prefetched at a time. Maybe we can make that more clear?

addyosmani:
My two cents are we shouldn't also apply `limit` to `options.urls`. Providing a single option that controls two related (..but slightly different) things feels like it could confuse folks.

src/index.mjs
lukeed: Microbundle doesn't do that (yet?)

I'm so proud 🤧

addyosmani:
Hey @jadjoubran. After discussing this implementation with Luke, we think there are a number of downsides to limiting by elements alone (and this can be solved using the new `ignores` feature). Something like..

```js
let count = 0;
quicklink({
  // Allows first 5 to run
  ignores: () => count++ < 5
})
```

Instead, we would like to go for limiting by outgoing network requests (through the prefetcher). So `limit` is passed through as an option the prefetcher then handles. We think the best way to do this is via a queue system that pulls 5 (or whatever the `limit` is) URLs from the toPrefetch group at a time. We then try to grab the next group after the first 5/`limit` resolve.

@lukeed is up for prototyping a version of this. Perhaps you'd be interested in refactoring this PR to carry that through to completion? Sorry once again for the back and forth, but I think we have a better sense for what this feature should look like now.

lukeed:
@jadjoubran I'm up for taking a first pass at it if you'd prefer, or I can review your continued work here. Up to you! Thanks for the work (and patience) so far! 😄

jadjoubran:
Makes sense!
And no worries, I'm learning from this and enjoying it!
@lukeed it's midnight here, so I'll leave the first prototype to you

addyosmani:
As we haven't seen much activity on this PR since last year, I'm going to close it up. Just wanted to say a big thank you for the enthusiasm exploring this initially @jadjoubran. If there are further changes related to this we want to explore, I might suggest we pick it up in a new issue/PR as there have been quite a few other changes to the repo since :)

Issue #31, opened 2018-12-13
Prefetch links from same domain?


maxigimenez:
Maybe set a guard to prefetch urls with the same domain? 
Because now either XHR or Fetch will fail for external links

naeluh:
This worked for me but yes I agree
```javascript
 let links = [...document.getElementsByTagName('a')].filter(a => {
    return a.hostname === window.location.hostname;
  });
  quicklink({urls:links});
```

addyosmani:
Reviewing some of the feedback from Twitter, quite a few folks have asked if there's a way to guard prefetching to same origin. That's a pretty fair request. 

I can see a few ways this could be done..

1. Same-origin is the default for all forms of prefetching (rel=prefetch, fetch(), XHR etc)
2. Same-origin guarding is an option (e.g `quicklinks({ sameDomainOnly: true })`)
3. Same-origin guarding is a documented pattern in the README/FAQ (similar to @naeluh's POC)

@lukeed would also love your take on this if you have thoughts on defaults :)

lukeed:
I think `whitelist` or `allowed` option would cover most use cases. It'd be an array of all domain strings the `quicklink` is permitted to fetch. We can ask users to provide their own domain in this list, or add an _additional_ option (`sameDomainOnly`) that pushes into that list.

```js
let allowed = [];
const observers = new IO(...)

export default function (opts) {
  opts = Object.assign({
    // ...
    whitelist: [],
  }, opts);

  let allowed = opts.sameDomainOnly ? [location.hostname] : opts.whitelist;

  // .... we should trust the options.urls given

  } else {
      Array.from(options.el.querySelectorAll('a'), link => {
        observer.observe(link);
        // <a> tags already parse URL bodies for us~!
        // empty "allowed" ~> everything OK
        let isAllowed = !allowed.length || allowed.includes(link.hostname);
        isAllowed && toPrefetch.add(link.href);
      });
  }
}
```

lukeed:
Oops, sorry @Avi98! 🙈 Just pulled the above into a PR

addyosmani:
Thanks, @lukeed! I'm in favor of a `whitelist` option with `sameDomainOnly` piggy-backing off that implementation.  I see you've already filed a PR with your POC above :)

[...]

PR #32, opened 2018-12-13
feat(restrict to same origins, add ignores)

lukeed:
Pulled my snippet from #31 into a PR.

The `origins` option is a list of permitted hostnames. The `sameOrigin` option indicates whether or not the domain's origin is all that's allowed. 

The use of `origins` and `sameOrigin` are mutually exclusive. If one has additional origins to allow, they should include their own within the larger `origins` array.

Adds 47 bytes in total.

---

_Closes #31, #33, and #38_

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -50,7 +50,7 @@ function prefetcher(url) {
 function isIgnored(node, filter) {
   return Array.isArray(filter)
     ? filter.some(x => isIgnored(node, x))
-    : (filter.test || filter)(node.href, node);
+    : (filter.test || filter).call(filter, node.href, node);
 }
 
 /**
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -39,17 +39,17 @@ function prefetcher(url) {
 }
 
 /**
- * Determine if the anchor tag can be prefetched.
+ * Determine if the anchor tag should be prefetched.
  * A filter can be a RegExp, Function, or Array of both.
  *   - Function receives `node.href, node` arguments
  *   - RegExp receives `node.href` only (the full URL)
  * @param  {Element}  node    The anchor (<a>) tag.
  * @param  {Mixed}    filter  The custom filter(s)
- * @return {Boolean}          If true, then it was allowed.
+ * @return {Boolean}          If true, then it should be ignored
  */
-function isAllowed(node, filter) {
+function isIgnored(node, filter) {
   return Array.isArray(filter)
-    ? filter.every(x => isAllowed(node, x))
+    ? filter.some(x => isIgnored(node, x))
     : (filter.test || filter)(node.href, node);
 }
 
@@ -64,7 +64,7 @@ function isAllowed(node, filter) {
  * @param {Object} options.el - DOM element to prefetch in-viewport links of
  * @param {Boolean} options.priority - Attempt higher priority fetch (low or high)
  * @param {Array} options.origins - Allowed origins to prefetch (empty allows all)
- * @param {Array|RegExp|Function} options.filter - Custom filter(s) that run after origin checks
+ * @param {Array|RegExp|Function} options.ignores - Custom filter(s) that run after origin checks
  * @param {Number} options.timeout - Timeout after which prefetching will occur
  * @param {Function} options.timeoutFn - Custom timeout function
  */
@@ -79,7 +79,7 @@ export default function (options) {
   observer.priority = options.priority;
 
   const allowed = options.origins || [location.hostname];
-  const filter = options.filter;
+  const ignores = options.ignores || [];
 
   options.timeoutFn(() => {
     // If URLs are given, prefetch them.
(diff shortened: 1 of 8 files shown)


addyosmani:
User-feedback: https://mobile.twitter.com/mxbck/status/1073334526313721863 :)

addyosmani:
User-feedback: https://mobile.twitter.com/mxbck/status/1073334526313721863 :) Thanks for being byte-conscious with this PR. I'm awaiting `bundlesize` in CI being approved for this repo.

src/index.mjs
```diff
@@ -62,7 +63,7 @@ export default function (options) {
 
   observer.priority = options.priority;
 
-  const allowed = options.origins || [];
+  const allowed = options.sameOrigin ? [location.hostname] : options.origins || [];
```
addyosmani: We might want to make it clear (perhaps in the jsdoc) what the default behavior for `sameOrigin` is? :)

* Benefit of `sameOrigin` defaulting to true: avoids issues with CORB, fetch(), XHR cross-domain
* Benefit of it defaulting to false: still get the benefits of rel=prefetch'ing third-party links by default (this early on, it's hard to get a feel for how much sites benefit from this in practice).

addyosmani:
Once we're happy with the defaults, the two things we'll probably want before landing this are tests and a README update for `origins` and `sameOrigin`. Happy to add those to this PR or defer to you, @lukeed. Any preferences?

src/index.mjs
lukeed: Sure! I can make a pass at the readme – I figured you'd drop some knowledge there anyway. 🎓 

This PR was just meant to match the current behavior. Technically, swapping `sameOrigin` to true would be a breaking change, I think. I'm generally hesitant towards major bumps so early on... but that's just me! 

As far as which way quicklink _should_ lean? No strong feelings, but if we are to keep both options, it's more config to disable sameOrigin restriction.


```js
// When sameorigin defaults true

// To turn off:
prefetch({
 origins: ['my', 'list', 'here'],
 sameOrigin: false
});
```

Another version (and maybe better?) is to work within a single `origin` option, which defaults to the sameOrigin initially. To turn _this_ off, you just re-define new `origin` list:

```js
// quicklink.js
opts = Object.assign({
  origin: [location.hostname]
}, opts);


// my-app.js
// Turn off sameOrigin
quicklink({ origin:false });
// or: { origin:[] }

// Defining custom origins also requires own origin be included:
quicklink({
  origin: [
    'my-app.com',
    'other-app.net'
  ]
})
```

lukeed: So, after https://github.com/GoogleChromeLabs/quicklink/issues/33#issuecomment-447174379, I think my vote is with this second option, along with a (breaking) change to restrict to same origin only.

This would mean `sameOrigin` is redundant & therefore removed. A combination of `origins` and `filter` should provide full control/customization – with `origins` exact-matching hostnames & `filter` handling the finer details. Similarly, one could opt-out of `origins` and move that work into their `filter` function(s) directly.

As suggested in that issue, when both `origins` and `filter` exist, the filter(s) should run _after_ the origin checks.

addyosmani: After reviewing the discussion in #33, I'm strongly inclined to agree with the second option. Users appear to want control over `origins`, allow and disallow lists (which a  `filter` pretty much provides too). I'm happy for us to proceed with your suggested model here, @lukeed (thanks a lot for thinking through this too..).

A change here would fix both #33 and #31.

addyosmani:
I'm going to hold off on changes on top of the latest commits as it appears we're aligned on shifting to `origins` and `filter` 👍

lukeed:
Sounds good! Will tackle it after #35 to skip avoid merge conflict noise

lukeed:
(Force pushed after rebase)

I ended up swapping to "ignores" at the last moment because I had confusion while running the tests. It wasn't clear to me if, when multiple filters were defined, a URL needed to pass _all_ filters or just one. I think the behavior is clear when there's only one RegExp or one Function ("keep anything that matches this") but not so clear when juggling multiple.

So, I swapped for "ignores" because that, to me, is crystal clear in that anything that matches an entry should be tossed. Everything we described above w/ `options.filter` can also be done with `options.ignores` but of course I can revert if `filter` has a clearly defined behavior in your mind.

Either way, I made it such that Function-types receive the URL _and_ the anchor tag itself, as requested in #38. It came at no extra cost really (literally 1 byte) & I can imagine it being useful.

Final size is 782B gzip, which is up from 672B after the latest round of golf.

test/test-ignore-multiple.html
```diff
@@ -0,0 +1,32 @@
...
+    <a href="4.html" style="position:absolute;margin-top:900px;">Link 4</a>
+    <script src="../dist/quicklink.umd.js"></script>
+    <script>
+        quicklink({
+            origins: true, // allow all initially
+            ignores: [
+                /example/,
+                (uri) => uri.includes('foobar'),
+                (uri, elem) => elem.textContent.includes('Spinner')
+            ]
```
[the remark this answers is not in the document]
lukeed: kudos to @aolko for the idea

addyosmani:
Let's land this. Thanks for working through a few different designs before landing on something that should really prove useful for folks ❤️

PR #34, opened 2018-12-14
convert the return value of Date.now() to seconds since that should be the default behaviour of a timestamp since unix epoch

zombieleet:


zombieleet:
I signed it!

PR #35, opened 2018-12-14
Golfing, Pt 2

lukeed:
Additional 🏌️ which can be reverted individually as requested.

Brings the library from 741B to 672B with no breaking changes this time~

---

Half-expecting the last commit to be reverted 😆

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -52,20 +52,25 @@ function prefetcher(url) {
  * @param {function} options.timeoutFn - Custom timeout function
  */
 export default function (options) {
-  options = options || {};
+  options = Object.assign({
+    timeout: 2e3,
+    priority: false,
+    timeoutFn: requestIdleCallback,
+    el: document,
+  }, options);
 
-  observer.priority = !!options.priority;
+  observer.priority = options.priority;
 
-  (options.timeoutFn || requestIdleCallback)(() => {
+  options.timeoutFn(() => {
     // If URLs are given, prefetch them.
     if (options.urls) {
       options.urls.forEach(prefetcher);
     } else {
       // If not, find all links and use IntersectionObserver.
-      Array.from((options.el || document).querySelectorAll('a'), link => {
+      Array.from(options.el.querySelectorAll('a'), link => {
         observer.observe(link);
         toPrefetch.add(link.href);
       });
     }
-  }, {timeout: options.timeout || 2e3});
+  }, {timeout: options.timeout});
 }
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -54,12 +54,11 @@ function prefetcher(url) {
 export default function (options) {
   options = Object.assign({
     timeout: 2e3,
-    priority: false,
     timeoutFn: requestIdleCallback,
     el: document,
   }, options);
 
-  observer.priority = options.priority;
+  observer.priority = !!options.priority;
 
   options.timeoutFn(() => {
     // If URLs are given, prefetch them.
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -52,24 +52,20 @@ function prefetcher(url) {
  * @param {function} options.timeoutFn - Custom timeout function
  */
 export default function (options) {
-  options = Object.assign({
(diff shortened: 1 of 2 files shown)


src/prefetch.mjs
```diff
@@ -94,7 +85,7 @@ function highPriFetchStrategy(url) {
     : fetch(url, {credentials: `include`});
 }
 
-const supportedPrefetchStrategy = support(`prefetch`)
+const supportedPrefetchStrategy = typeof document !== 'undefined' && support('prefetch')
```
lukeed: To be honest, I don't think we even need the `typeof` check here since everything else already assumes/requires that we be in a browser context. Even with this, a server-side context would have to provide global shims and/or wrap the require statement in additional `typeof` checks.

src/prefetch.mjs
```diff
@@ -104,20 +95,14 @@ const supportedPrefetchStrategy = support(`prefetch`)
...
-    if ((navigator.connection.effectiveType || '').includes('2g')) {
-      return;
-    }
-    // Don't prefetch if Save-Data is enabled...
-    if (navigator.connection.saveData) {
-      return;
-    }
+  if (conn = navigator.connection) {
+    // Don't prefetch if the user is on 2G. or if Save-Data is enabled..
+    if ((conn.effectiveType || '').includes('2g') || conn.saveData) return;
```
lukeed: I know you asked for this to be reverted in the last attempt, but perhaps the `conn` abbreviation preserves the clarity. It did for me.

src/prefetch.mjs
addyosmani: I'm okay with dropping the `typeof` check here given the above context.

src/prefetch.mjs
addyosmani: I'm okay with us switching up to this abbreviation. I agree it reads a little better than the last version 👍

src/index.mjs
```diff
@@ -52,24 +52,20 @@ function prefetcher(url) {
  * @param {function} options.timeoutFn - Custom timeout function
  */
 export default function (options) {
-  options = Object.assign({
-    timeout: 2e3,
-    timeoutFn: requestIdleCallback,
-    el: document,
-  }, options);
+  options = options || {};
```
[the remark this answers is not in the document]
lukeed: 😎 it's cool hahah

src/index.mjs
```diff
@@ -52,20 +52,25 @@ function prefetcher(url) {
...
-  options = options || {};
+  options = Object.assign({
+    timeout: 2e3,
+    priority: false,
+    timeoutFn: requestIdleCallback,
+    el: document,
+  }, options);
 
-  observer.priority = !!options.priority;
+  observer.priority = options.priority;
```
addyosmani: Did we mean to drop https://github.com/GoogleChromeLabs/quicklink/blob/c578c32c1c511f7cef7c1859fc480140f1e3ceb0/src/index.mjs#L62 as priority is back in the Object.assign block? :)

lukeed: This one writes into the Observer instance, which is referenced within `prefetcher()` – it's just a ploy to move the option value out of the function.

Here I just removed the `!!` which was used to cast undefined -> false.

---

Checking on `document.head` within Edge for ya

addyosmani: Thanks! Appreciate you checking :)

src/prefetch.mjs
```diff
@@ -52,7 +43,7 @@ function linkPrefetchStrategy(url) {
     link.onload = resolve;
     link.onerror = reject;
 
-    (document.head || document.querySelector(`script`).parentNode).appendChild(link);
+    document.head.appendChild(link);
```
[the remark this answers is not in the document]
43081j: FYI the dom types shipped in TS specify it as being optional and they're generally generated from the IDL taken from the spec. so maybe browsers just ignored that part and always create it?

src/index.mjs
addyosmani: Fantastic. Thanks for verifying, Luke!

PR #37, opened 2018-12-14
feat(index.mjs): normalize URLs being prefetched (fixes #27)

addyosmani:
reviewers proposed: @lukeed @mathiasbynens

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -35,7 +35,7 @@ const observer = new IntersectionObserver(entries => {
  */
 function prefetcher(url) {
   toPrefetch.delete(url);
-  prefetch(new URL(url).toString(), observer.priority);
+  prefetch(new URL(url, location.href).toString(), observer.priority);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


addyosmani:
Luke also proposed the `prefetch(new URL(url), priority)` abbreviation.

addyosmani:
One fyi downside to this change is that while previously "loose" paths would prefetch without issue (['2.html', '4.html']), at least based on our tests, fully qualified paths are now required (`['http://127.0.0.1:8080/test/2.html','http://127.0.0.1:8080/test/4.html']`).

lukeed:
We could do `new URL(url, location.href)`, which builds a URL string w/ 2nd arg as reference

addyosmani:
> We could do new URL(url, location.href), which builds a URL string w/ 2nd arg as reference

Great idea. This allowed reverting back to the old tests and appears to have done the trick. I'll hold off on landing this to give @mathiasbynens a chance to sanity check what we're doing here.

mathiasbynens:
LGTM!

addyosmani:
Thanks folks!

PR #41, opened 2018-12-14
introduce a filter option

### LINKED ISSUE #38: Ignore elements?
Can you ignore specific elements? "Quicklink everything but [...]" case

43081j:
Fixes #38 

I gather its possible you already began work on this, so feel free to close this if that's the case.

Also FYI there's some bug in microbundle 0.7, you should upgrade it as the build currently fails locally (due to microbundle missing a package, the bug).

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -50,6 +50,7 @@ function prefetcher(url) {
  * @param {Boolean} options.priority - Attempt higher priority fetch (low or high)
  * @param {Number} options.timeout - Timeout after which prefetching will occur
  * @param {function} options.timeoutFn - Custom timeout function
+ * @param {function} options.filter - Custom filter function
  */
 export default function (options) {
   options = Object.assign({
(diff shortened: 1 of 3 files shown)


src/index.mjs
```diff
@@ -68,8 +69,10 @@ export default function (options) {
     } else {
       // If not, find all links and use IntersectionObserver.
       Array.from(options.el.querySelectorAll('a'), link => {
-        observer.observe(link);
-        toPrefetch.add(link.href);
+        if (!options.filter || options.filter(link.href)) {
```
43081j: alternatively we could default to `() => true` so this becomes `Array.from(blah).filter(filter).forEach(...)`

lukeed:
Definitely an option 👍 The only thing I'd like to ask if we were going to allow filters to be RegExp, Function, and/or Array<RegExp|Function>.

I'll push up my impl on other branch & we can discuss each. Happy to build from here tho

addyosmani:
Thanks for working on a version of filtering, @43081j. After discussing a few designs with @lukeed we landed on a solution for filtering (`ignores`) and same-domain limiting (`origins`) which we think capture most of the use-cases folks will want to be solved by a filtering solution.

This is now landed with docs in `quicklink` 1.0.0. We're going to go ahead and close this PR but super appreciate your time on it.

Issue #39, opened 2018-12-14
Does not work as a traditional JS resource – Uncaught ReferenceError: module is not defined


olpeh:
Hello, this looks like a very handy library. However, I'm having trouble using it **without** using it as an ES-module.

**My source code:**
```
<script src="https://unpkg.com/quicklink@0.1.2/dist/quicklink.js"></script>
<script>
  quicklink();
</script>
```

**The output:** 
```
quicklink.js:1 Uncaught ReferenceError: module is not defined
    at quicklink.js:1
(anonymous) @ quicklink.js:1
(index):142 Uncaught ReferenceError: quicklink is not defined
    at (index):142
```

**Expected result**
I would expect it to work since the readme says so.

lukeed:
Hey, for browser/UMD usage thru unpkg.com, you should use this asset: `https://unpkg.com/quicklink@0.1.2/dist/quicklink.umd.js`

The one you requested is the Node/CommonJS version

olpeh:
Oh, okay thanks, that works. Maybe that could be written in the readme?

lukeed:
For sure, will do – thanks!

HiFaraz:
@lukeed the README current advises to use `<script src="dist/quicklink.js"></script>`. Should that also be changed to `<script src="dist/quicklink.umd.js"></script>`? Doing that fixed a similar error in my app. Happy to submit a PR for this :)

lukeed:
@HiFaraz Oh, yes! That file is for CommonJS. Would you like to do a PR for the README fixes? Everything inside a `<script>` tag should be the UMD file – except the example(s) with ESM imports.

HeikoMamerow:
Sorry for silly OT question, but why did you get the common js file this "umd"?

HiFaraz:
@lukeed I'll submit a PR in the next couple of days.

PR #42, opened 2018-12-14
unpkg link to umd version to avoid issue #39

Antoinebr:
The unpkg.com link from the documentation is not working because the unpkg link redirects to ```/dist/quicklink.js``` instead of ```/dist/quicklink.umd.js``` as mentioned by #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ For use with [node](http://nodejs.org) and [npm](https://npmjs.com):
 npm install --save quicklink
 ```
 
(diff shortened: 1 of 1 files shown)


lukeed:
Thank you, but this was fixed in #40 just recently. It will be reflected in the next version & start redirecting correctly to the UMD file.

addyosmani:
Thanks for sending this through. We've addressed this in #40 and will be publishing a new release in the coming days.

PR #48, opened 2018-12-15
[FIX #14] Add config for IO/initialize in default function

### LINKED ISSUE #14: Consider adding `rootMargin` to the Intersection Observer
Feedback from @surma:

> I feel like by default you should add a rootMargin to the IO to trigger the load before something comes into the viewport. 

So, `rootMargin` defaults to "0px 0px 0px 0px", meaning the intersection will be computed between the root element's unmodified bounds rectangle and the target's bounds. This change would be made to https://github.com/GoogleChromeLabs/quicklink/blob/master/src/index.mjs#L21

aretheregods:
Fixes #14
How about this?

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -64,8 +64,6 @@ function isIgnored(node, filter) {
  * @param {Element} options.config.root - The root container in which elements are observed
  * @param {string} options.config.rootMargin - Margin added to root before intersection observation testing
  * @param {number} options.config.threshold - Percentage of observed element in root before IO callback is called
- * @param {number} options.config.threshold - percentage of observed element in root before IO callback is called
- * @param {number} options.config.threshold - Percentage of observed element in root before IO callback is called
  */
 export default function (options) {
   options = Object.assign({
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -18,18 +18,8 @@ import prefetch from './prefetch.mjs';
 import requestIdleCallback from './request-idle-callback.mjs';
 
 const toPrefetch = new Set();
-
-const observer = new IntersectionObserver(entries => {
-  entries.forEach(entry => {
-    if (entry.isIntersecting) {
-      const link = entry.target;
-      if (toPrefetch.has(link.href)) {
-        observer.unobserve(link);
-        prefetcher(link.href);
-      }
-    }
-  });
-});
+// Add observer to namespace for prefetcher
+let observer;
 
 /**
  * Prefetch a supplied URL. This will also remove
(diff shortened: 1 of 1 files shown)


src/index.mjs
```diff
@@ -19,14 +19,12 @@ import requestIdleCallback from './request-idle-callback.mjs';
...
 
-const observer = new IntersectionObserver(entries => {
-  entries.forEach(entry => {
-    if (entry.isIntersecting) {
-      const url = entry.target.href;
-      if (toPrefetch.has(url)) prefetcher(url);
-    }
-  });
-});
+const observerConfig = {
```
lukeed: Let's remove this – as is, this just redefines the existing defaults without a way to customize it.

src/index.mjs
```diff
@@ -75,6 +73,16 @@ export default function (options) {
...
+  
+  // actually initialize intersection observer here
+  observer = new IntersectionObserver(entries => {
+    entries.forEach(entry => {
+      if (entry.isIntersecting) {
+        const url = entry.target.href;
+        if (toPrefetch.has(url)) prefetcher(url);
+      }
+    });
+  }, observerConfig);
```
lukeed: Let's replace this with a `config` object that allows the user to directly manipulate the IO settings. 

```suggestion
  }, options.config || {});
```

The only drawback here is that it requires the user to know _what_ those options are... but we can easily link to [its options](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#Intersection_observer_options) in the docs IMO. Plus, I personally prefer to expose native options, rather than reinvent them with new names.

What do you think @addyosmani ?

aretheregods: Yeah, I was gonna do that. But I didn't wanna step add something to options without advisement. I'll make that change.

aretheregods:
I removed observerConfig, but still need to add the exposed options for options.config.

src/index.mjs
addyosmani: I just wanted to add that I'm in support of exposing native names vs defining our own variants where possible 👍

Thanks for working on this, @aretheregods!

addyosmani:
As a passing note, it looks like this will need to be rebased against master ♻️

aretheregods:
I believe that rebasing business might possibly be resolved...

addyosmani:
@lukeed I believe this PR managed to address most of our feedback. Was there anything else remaining before you were happy to sign off on it?

@aretheregods sorry to ask again - would you mind rebasing one more time? It's entirely my fault for not circling back to wrap up reviews when you first tried addressing our feedback. Would love to wrap this one up.

src/index.mjs
```diff
@@ -70,14 +60,29 @@ function isIgnored(node, filter) {
...
+ * @param {string} options.config.rootMargin - Margin added to root before intersection observation testing
+ * @param {number} options.config.threshold - Percentage of observed element in root before IO callback is called
  */
 export default function (options) {
   options = Object.assign({
     timeout: 2e3,
     priority: false,
     timeoutFn: requestIdleCallback,
     el: document,
+    config: {}
```
lukeed: We don't need this line, as `new IntersectionObserver(noop, undefined)` is still valid 😆 
```suggestion
```

addyosmani:
@aretheregods a friendly ping. Looks like we are close to this being landable :)

PR #49, opened 2018-12-15
fix: prevent if link is URL fragment

arkist:
Currently, quicklink trying prefetch even if link is just plain URL fragment.

```html
<link rel="prefetch" href="http://localhost:8080/demos/basic.html#3rdScreen">
```


lukeed:
My gut reaction is that this should live within user `filters` 🤔 While it makes sense in the general case, there'd be no way around this if your site's client was using the `/#/` (or similar) routing convention. For example:

```js
// localhost:8000/#/hello
location.pathname = '/';
location.hash = '#/hello';

// localhost:8000/#/world
location.pathname = '/';
location.hash = '#/world';
```

Here, the hostname & pathnames are identical, but the unique hashes _can actually_ be different pages – whether they're rendered dynamically thru Angular/Vue/Preact/etc. In this case, external prefetching is & should still be enabled.

Including this logic into the core of `quicklink` would block this pattern.

addyosmani:
I'm tempted to agree with @lukeed that control over URL fragment behavior is something we should defer to the new `ignores` feature (in latest stable). It's possible to filter out any links in a page with fragments and only prefetch those that don't have them if that's the desired outcome. 

My preference here would be that we document in the README how to use `ignores` to achieve the above, but not make this change to core. </two-cents>

arkist:
@addyosmani Okay I'll try later 👌 thanks! 😄

arkist:
@lukeed woah, you're right 🤯 as you explained, my code could be a problem in that kind of querystring case.
thank you again!

arkist:
my last thought about treat URL fragment is `/a` to `/b#step2`, `/b#step3` case. 
in that case you should check `pathname` not just check hash included.
also, should store `preFetched` url *without* hash maybe 🤔

arkist:
Oh, I pressed the 'close PR' button by mistake 😅 
(but it's okay, isn't it? 😉
if you don't mind,

Issue #43, opened 2018-12-14
Unobserve prefetched elements


lfre:
Depending on the number of links on a page, or if new links are added programmatically after the initial call. Both the query selector in the timeout method https://github.com/GoogleChromeLabs/quicklink/blob/master/src/index.mjs#L70, and/or the entries loop in the observer callback https://github.com/GoogleChromeLabs/quicklink/blob/master/src/index.mjs#L23, could be problematic for subsequent calls.

I think it would be worth to `unobserve` elements that are already prefetched.

Additionally, instead of requiring additional calls to `quicklink({ urls: [..]})`, which I think it is still needed to override the intersection observer; a `mutationobserver.mjs` could be enabled through an option, or to keep the lib size down, expose it in dist and allow its default export to be passed as an option for runtime.

Let me know what you think and I can tackle these. 👍 
Thanks!

lukeed:
I think this can be enabled by default, without any options. It can actually be done with 2 lines of code & It would serve to make the observer more efficient.

Currently, we immediately select all links within the `document` (default) and observe them. These links are repeatedly re-selected & re-assigned to the `observer`. The items that prevent `quicklink` from dispatching multiple requests to the same URL are:

* `Set` natively retaining unique items – we can (and are) pushing the same `link.href` into the set on every `rIC` loop
* the `preFetch={}` cache tracking URLs & blocking repeat requests

So, while these _are_ doing their jobs correctly, they are doing more work than they have to.

---

Put differently (I tend to ramble, sorry) `quicklink` will continue to observe & track the same links over & over again on a static page. Everything except the final `prefetch()` call behaves the same, indefinitely. Ideally, `quicklink` selects all links on its interval & only processes the new items, if any.

I can PR this if we're all in agreement. It really should only be 2 LoC / ~20 bytes

lukeed:
Perhaps we should rename this issue (or open a new one) for supporting infinite scroll? Now that quicklink is now _technically_ unobserving processed elements, this issue has sparked a larger feature.

lfre:
Thanks, @lukeed @addyosmani! I split the specific issue [here](https://github.com/GoogleChromeLabs/quicklink/issues/54). Apologies for lumping them together, infinite scroll is where the issue was most noticeable.

Closing this.

PR #51, opened 2018-12-16
Fix: Unobserve links once prefetched

lukeed:
- saved reference to link, not url, for better gzip
- 787 gz / 654 br

---

_Closes #43_


addyosmani:
You're entirely right that we can unobserve individual links once prefetched with minimal change. I'm happy to land this. 

I think #43 may morph into a larger feature. I'm interested in how we can support more complex use cases (like infinite scroll). Let's ship this.

renosbest:
Confused here im I a part of anything done all I do is run test on terminals right now just simple imputs

PR #56, opened 2018-12-18
feat: add limit number of network option

HOUCe:
Implement limit number of network option to ensure the maximum network request


HOUCe:
This feature is what I need for my project cause I am wondering fetching every link will increase huge QPS suddenly or adding `<link>` tags in such short time is ok or not.

But whether is ideal / intended for this project, I will leave you guys decide (anyway is a small addition as you said).

addyosmani:
Thanks a lot for sharing your implementation, @HOUCe. 

I think we're after a strategy that can factor in controlling and limiting what is prefetched beyond init. We're going to continue evolving this as part of #23 but appreciate your PR nevertheless 😄

PR #60, opened 2018-12-19
Update repo path in package.json

coliff:
change from:
`https://github.com/addyosmani/quicklink.git`
to:
`https://github.com/GoogleChromeLabs/quicklink.git`


addyosmani:
Thanks!

PR #65, opened 2018-12-20
ci: update Yarn with the recommended official commands

DanielRuf:


.travis.yml
```diff
@@ -9,11 +9,14 @@ cache:
     - node_modules
 
 before_install:
+  - sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
```
addyosmani: Thanks for the PR! For posterity, is the official reference you were noting this one? https://yarnpkg.com/lang/en/docs/install-ci/

DanielRuf: Exactly, it is https://yarnpkg.com/lang/en/docs/install-ci/#travis

I should have mentioned this.

PR #68, opened 2018-12-23
fix(README): use UMD file for <script> tags

HiFaraz:
Browsers should use the UMD file (unless importing as ES modules).

ref: https://github.com/GoogleChromeLabs/quicklink/issues/39

cc: @lukeed


HiFaraz:
I signed the CLA.

Issue #59, opened 2018-12-18
Uncaught (in promise) 


MatTeague:
Hi,

I've included the umd version of quicklinks but I keep getting several console errors repeating the error:

```
Uncaught (in promise) Event {isTrusted: true, type: "error", target: link, currentTarget: null, eventPhase: 0, …}bubbles: falsecancelBubble: falsecancelable: falsecomposed: falsecurrentTarget: nulldefaultPrevented: falseeventPhase: 0isTrusted: truepath: (5) [link, head, html.js.no-focus-outline.sizes.customelements.history.pointerevents.postmessage.webgl.websockets.css…, document, Window]returnValue: truesrcElement: linktarget: linktimeStamp: 2088.500000001659type: "error"__proto__: Event
Promise.then (async)
r @ quicklink.umd.js:1
c @ quicklink.umd.js:1
(anonymous) @ quicklink.umd.js:1
(anonymous) @ quicklink.umd.js:1
```

Any advice with resolving these errors would be great. Thanks.

addyosmani:
Could you share more details about your testing setup so we can try reproducing the error? (e.g browser, any configuration options passed, minimal test case). Cheers!

lukeed:
Looks like the `linkPrefetchStrategy` method was used, which likely means that the `link.onerror` triggered. Maybe we should include a simple logger on `catch()` here? https://github.com/GoogleChromeLabs/quicklink/blob/master/src/prefetch.mjs#L109

Functionally, it wouldn't be very helpful, but it's a shot at debugging.

Or we can expose an option for user-derived error handler

lukeed:
Hey @MatTeague, It looks like that original error you reported is throwing because `/contact-us` does not exist. Because of this line (https://github.com/GoogleChromeLabs/quicklink/blob/master/src/prefetch.mjs#L44), any non-OK responses will just emit upwards, which is why an `Event` instance is logged to the console.

As for your latest report, `quicklink` won't be defined because your script tag has a `defer` attribute on it:

```html
<script src="/assets/js/quicklink.umd.js" defer></script>
```

This will execute the `<script id="quicklink">` block first, since it's not (and can't) be deferred. The `load` and `DOMContentLoaded` events fire after all script sources (deferred or not) are ready.

tutuzhangdeqingtian:
Hello @MatTeague ,have you solve the original error? What do you do to solve it? I meet the same error. I want to learn from you about resolving this error.

[...]

PR #70, opened 2018-12-27
fix:Uncaught (in promise)

ChuanfengZhang:
May not be a good solution, but works:)
https://github.com/GoogleChromeLabs/quicklink/issues/59#issuecomment-448353766
ref: #59 
cc: @lukeed

### DIFF
--- a/src/prefetch.mjs
+++ b/src/prefetch.mjs
@@ -109,6 +109,8 @@ function prefetcher(url, isPriority, conn) {
   // Wanna do something on catch()?
   return (isPriority ? highPriFetchStrategy : supportedPrefetchStrategy)(url).then(() => {
     preFetched[url] = true;
+  }, (err) => {
+    // Wanna do something?
   });
 };
 


addyosmani:
@lukeed do you know if this one is still needed for the uncaught error?

lukeed:
This will catch them, but it'll be silent to the console _and_ to the developer.

Maybe the ideal is to make quicklink accept an error handler, which gets passed into our [`prefetcher` wrapper](https://github.com/GoogleChromeLabs/quicklink/blob/master/src/index.mjs#L39-L42), else noop.

That way, the `prefetch` itself has no built-in silencing an those who load it directly don't fight against it.

Something like this?

```js

const noop = () => {};

// main
  if (!options) options = {};
  
  // Attach ref to `toPrefetch` set
  toPrefetch.fetch = url => {
    toPrefetch.delete(url);
    prefetch(new URL(url, location.href).toString(), observer.priority).catch(options.onError || noop);
  };
  
  // ...
}

const observer = new IntersectionObserver(entries => {
  entries.forEach(entry => {
    if (entry.isIntersecting) {
      const link = entry.target;
      if (toPrefetch.has(link.href)) {
        observer.unobserve(link);
        // -- prefetcher(link.href);
        toPrefetch.fetch(link.href);
      }
    }
  });
});
```

addyosmani:
Hey folks. There hasn't been much further movement on this issue since April. With this in mind, @ChuanfengZhang did you have thoughts on adjusting your proposed solution per the snippet @lukeed was kind enough to share? 

Thumbs up on the original issue in #59 appear to suggest a few users have run into this. Would love to gauge interest in someone completing this patch (or us closing this PR if we don't think we'll be heading down this path further).

addyosmani:
Closing this PR for now. Please feel free to open up a new issue if there's interest in picking this back up :)

PR #73, opened 2019-01-07
Update microbundle to fix "missing JSX plugin" issue

developit:
There are no breaking changes from 0.7 -> 0.9. Size remains the same.


addyosmani:
👍

PR #74, opened 2019-01-07
Size optimizations

developit:
This PR introduces a few quick size optimizations, mainly inlining the `options` handling to avoid the call to `Object.assign()` (instead applying defaults throughout the exported function).

One thought I had was that hoisting these defaults into constants would be free (they get inlined by microbundle/terser) and bring back some of the visibility lost from removing the assign().

/cc @lukeed


addyosmani:
These byte-savings are great. Thanks, @developit!

PR #98, opened 2019-04-29
Publish dist directory

styfle:
See package.json docs https://docs.npmjs.com/files/package.json#files


lukeed:
We can't currently because in order to access `prefetch` directly, we have to traverse `src/` contents

PR #103, opened 2019-06-30
added readme draft

anton-karlovskiy:
For now it's just readme draft.
After I confirmed some README convention and how to write it, I'll update it shortly.

### DIFF
--- a/demos/news-react-router/README.md
+++ b/demos/news-react-router/README.md
@@ -1,0 +1,22 @@
+## Lab: Quicklink
+This demo is based on Create-React-App and React-Router.
+
+## Dropping the Quicklink default code(inside window onload event) into public/index.html
+In SPA, window-onload-event is just called one time when the website is loaded or only when it's refreshed.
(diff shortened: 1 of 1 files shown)


addyosmani:
On page load, requests to prefetch /world, /tech, /contact are made by Quicklink. However, these all appear to be ~104 byte entries. I don't believe fetches to these links are actually loading the content for each route chunk... :)

Screen Shot 2019-07-03 at 5 31 08 PM

If you navigate to /world for example, 8.chunk.js is loaded and this is what contains the content for the route:

Screen Shot 2019-07-03 at 5 31 50 PM

I wonder if there's a way for us to map back links for individual chunks to the routes being prefetched so we could do something like... 

```js
quicklink({
   urls: ['/8.chunk.js', '/6.chunk.js'] etc
});
```

addyosmani:
Closing this as we are exploring the React solution elsewhere and will circle back on a demo in a few weeks.

PR #110, opened 2019-07-22
HTTPS link to nodejs.org

coliff:
safer and saves a HTTP redirect

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ This project aims to be a drop-in solution for sites to prefetch links based on 
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


addyosmani:
Thanks, @coliff! 

I think you may have asked what other areas of the project we could use help out with. I would say the issue tracker (e.g checking what we should probably tackle vs. what are questions better deferred to Stack Overflow for example) would be super valuable.

coliff:
Hi @addyosmani  - yep, sounds good. I'll go through those issues and see where I can help out. I opened a PR with a bunch of small improvements for the tests and demos too.

PR #116, opened 2019-08-02
Add homepage and bugs links to package.json

coliff:
Can be useful to have there here:
ref:
- https://docs.npmjs.com/files/package.json#homepage
- https://docs.npmjs.com/files/package.json#bugs


addyosmani:
Thanks!

PR #118, opened 2019-08-19
Chore: More Golfing~

lukeed:
While not much, it can't hurt 😇 I do have another (unpushed) commit that brings this down further; see below.

***Before:***

```
        804 B: quicklink.js.gz
        667 B: quicklink.js.br
        804 B: quicklink.mjs.gz
        668 B: quicklink.mjs.br
        877 B: quicklink.umd.js.gz
        729 B: quicklink.umd.js.br
```

***After:***

```
        795 B: quicklink.js.gz
        661 B: quicklink.js.br
        795 B: quicklink.mjs.gz
        663 B: quicklink.mjs.br
        870 B: quicklink.umd.js.gz
        725 B: quicklink.umd.js.br
```

***Possible:***

```
        788 B: quicklink.js.gz
        654 B: quicklink.js.br
        789 B: quicklink.mjs.gz
        657 B: quicklink.mjs.br
        864 B: quicklink.umd.js.gz
        721 B: quicklink.umd.js.br
```

The  gotcha here though is that it removes the `window.` prefixes on `IntersectionObserver` and `requestIdleCallback`. This means that browsers that don't support them will throw an Error instead of silently doing nothing. 

That's a big change, so didn't push it

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -74,7 +74,7 @@ function isIgnored(node, filter) {
 export default function (options) {
   if (!options) options = {};
 
-  observer && (observer.priority = options.priority || false);
+  if (observer) observer.priority = !!options.priority;
 
   const allowed = options.origins || [location.hostname];
   const ignores = options.ignores || [];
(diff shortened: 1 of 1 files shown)


lukeed:
TravisCI error is `Error: Failed to launch chrome!` 🤷‍♂

addyosmani:
This will be a fun one to debug ... :) 

I've restarted the build a few times but it seems to still be failing to launch Chrome. I haven't been able to determine the cause, but if it's some Travis variance, I'll try again in a while just in case.

PR #120, opened 2019-08-22
Major: Rework exports; Add `throttle` and `limit` options

### LINKED ISSUE #4: Option to limit the maximum number of prefetched requests
It would be nice to limit the number of links that will be prefetched
which can also default at a sane number.

Proposed API: `options.limit`

Would love to get your feedback on naming & default limit

### LINKED ISSUE #79: Consider allowing quicklink to fail softly if IntersectionObserver isn't available.
I just recently launched a new web site that uses quicklink. It's impressive, and improves the perception of speed to a point where a static website feels almost like an SPA at times. Good work!

One thing I noticed, though, is that quicklink crashes when Intersection Observer is not available. I'm aware that the documentation recommends a polyfill, but do you think it would make sense to fail softly when Intersection Observer is unavailable? Perhaps using [this check](https://github.com/w3c/IntersectionObserver/blob/master/polyfill/intersection-observer.js#L16-L18) and using `console.warn`?

In any event, great work on this. This script has tremendous utility in improving perceived performance, and I look forward to using it in future projects.

lukeed:
_Sorry for the big diff; I was zoning & not really paying attention to git._

***TLDR:***

- adds `limit` option – _total_ allowed requests (#4)
- adds `throttle` option – max concurrency control via [`throttles`](https://github.com/lukeed/throttles) (https://github.com/GoogleChromeLabs/quicklink/issues/4#issuecomment-446270693)
- removes `urls` option (temp?)
- exports `prefetch` method directly
- exports previous `default` as `listen` method
- aborts `listen` early if no IO support
- only require IO support when using `listen`
- when okay, `listen` returns a `reset` / unlisten function

***Code Size***

```
// Before
        804 B: quicklink.js.gz
        667 B: quicklink.js.br
        804 B: quicklink.mjs.gz
        668 B: quicklink.mjs.br
        877 B: quicklink.umd.js.gz
        729 B: quicklink.umd.js.br

// After
        871 B: quicklink.js.gz
        734 B: quicklink.js.br
        878 B: quicklink.mjs.gz
        741 B: quicklink.mjs.br
        949 B: quicklink.umd.js.gz
        798 B: quicklink.umd.js.br
```

***Notes:***

I realized we were using the `toPrefetch` set somewhat incorrectly. We [did a bunch of work](https://github.com/GoogleChromeLabs/quicklink/blob/master/src/index.mjs#L97) to figure out if we should "queue" the link's `href` value, and then, once the link was intersecting, [we did more work](https://github.com/GoogleChromeLabs/quicklink/blob/master/src/index.mjs#L26) to see if the `link.href` was approved, and then finally [did more work](https://github.com/GoogleChromeLabs/quicklink/blob/master/src/index.mjs#L40) to clear the queue.

That's a lot of work.

Instead, it was much simpler to do the _first_ set of work to see if we should observe the `link` in the first place! Then anything else the IO is processing is already okay'd for processing.

This leaves the `toPrefetch` set as a track record of what _has been_ called, which is identical to what [this makeshift cache](https://github.com/GoogleChromeLabs/quicklink/blob/master/src/prefetch.mjs#L19) was doing, too.

Thirdly, these individually-operating caches were individually operating, which means that _when_ you were brave enough to venture into `quicklink/src/prefetch.mjs`, you were prefetching things that (a) may have already been prefetched; and/or (b) prefetching things that your IO won't ever know about.

The best way to solve this was to get everything in the same file and operating on the same cache list. It also allows for easy access to the `prefetch` function w/o manually traversing the module contents https://github.com/GoogleChromeLabs/quicklink/pull/98#issuecomment-522693684 which actually prevented code reuse too (quicklink was bundled, but reached back in for another copy of source code).

Finally, I removed the `urls` option _because_ of the easily accessible `prefetch` export. It feels redundant and very out of place now that the "Go get this" and the "Please listen to that" behaviors have been separated. Of course, I can revert this if it should be kept.

***TODOs***

- [x] Add tests for new features (how to test throttle?)
- [x] Add/Update documentation

---

Closes #4
Closes #79
Closes #118 
Unblocks #98


src/index.mjs
```diff
@@ -63,40 +43,89 @@ function isIgnored(node, filter) {
...
-          isIgnored(link, ignores) || toPrefetch.add(link.href);
+  const observer = new IntersectionObserver(entries => {
+    entries.forEach(entry => {
+      if (entry.isIntersecting) {
+        observer.unobserve(entry = entry.target);
+        // Do not prefetch if will match/exceed limit
+        if (toPrefetch.size < limit) {
+          toAdd(() => {
+            // TODO: Don't need isPriority?
+            prefetch(entry.href, isPriority).then(isDone);
```
lukeed: Would be good to add a `catch` handler here now; exposed as `options.onError` perhaps? https://github.com/GoogleChromeLabs/quicklink/pull/70#issuecomment-482699527

This would still allow/require user to do custom error handling when calling `quicklink.prefetch` directly.

src/index.mjs
```diff
@@ -63,40 +43,89 @@ function isIgnored(node, filter) {
...
+export function listen(options) {
   if (!options) options = {};
+  if (!window.IntersectionObserver) return;
+
+  const [toAdd, isDone] = throttle(options.throttle || 1/0);
+  const limit = options.limit || 1/0;
 
-  observer && (observer.priority = options.priority || false);
+  // TODO: I think this isn't needed?
+  const isPriority = !!options.priority;
```
lukeed: I'm sure I just don't fully realize the purpose, but I don't think this is strictly necessary since anything that comes through `listen` passes thru the IO, which will only call `prefetch` once a link is visible. 

The only use I can see is when a user slices up their page(s) into smaller `options.el` containers, saying that ContainerA is high priority and the others aren't. Maybe that's a big one & makes it worth keeping 😆

src/index.mjs
```diff
@@ -63,40 +43,89 @@ function isIgnored(node, filter) {
...
+      if (!allowed.length || allowed.includes(link.hostname)) {
+        // If there are any filters, the link must not match any of them
+        isIgnored(link, ignores) || observer.observe(link);
+      }
+    });
+  }, {
+    timeout: options.timeout || 2e3
+  });
+
+  return function () {
```
lukeed: This is good for SPA environments, where `listen` is initially called on a parent container and then any page/route-change can dump the contents of what _was_ relevant to the previous page.

src/index.mjs
addyosmani: > Would be good to add a catch handler here now; exposed as options.onError perhaps?

SGTM and thank you for circling back on #70. Even though we would be deferring to users for custom error handling in `.prefetch()`, adding a `catch` -> onError is a good call here.

addyosmani:
> Finally, I removed the urls option because of the easily accessible prefetch export. It feels redundant and very out of place now that the "Go get this" and the "Please listen to that" behaviors have been separated. Of course, I can revert this if it should be kept.

I agree with the rationale for removal, but do have some questions. If we drop the `urls` option, users would need to iterate through a supplied array of URLs they want to prefetch manually, right? i.e similar to the current prefetch import behavior if using it as a standalone:

```
const urls = ['1.html', '2.html'];
const promises = urls.map(url => Quicklink.prefetch(url));
Promise.all(promises);
```
I would be supportive of dropping `urls` if we modified the prefetcher to accept either a single URL string or an array of URLs (to preserve ergonomics and ease upgrading for anyone switching to this version). Wdyt?

addyosmani:
Thank you for working on these changes, @lukeed! 

> The best way to solve this was to get everything in the same file and operating on the same cache list. It also allows for easy access to the prefetch function w/o manually traversing the module contents #98 (comment) which actually prevented code reuse too (quicklink was bundled, but reached back in for another copy of source code).

A shift to working against the same cache list was a good call. I'm also glad to hear you discovered a set of changes to improve code reuse when we've finally bundled Quicklink. 

> Sorry for the big diff; I was zoning & not really paying attention to git.

No worries. I'm happy to treat this PR as the next major release candidate once we are further along in the reviews. Given the breaking changes in the API surface, we should probably consider this Quicklink 2.0.0. Top of mind is making sure there's feature parity in this new version. Looking forward to the tests :)

src/index.mjs
```diff
@@ -63,40 +43,87 @@ function isIgnored(node, filter) {
...
+ * @param {Boolean} [options.priority] - Attempt higher priority fetch (low or high)
+ * @param {Array} [options.origins] - Allowed origins to prefetch (empty allows all)
+ * @param {Array|RegExp|Function} [options.ignores] - Custom filter(s) that run after origin checks
+ * @param {Number} [options.timeout] - Timeout after which prefetching will occur
+ * @param {Number} [options.throttle] - The concurrency limit for prefetching
+ * @param {Number} [options.limit] - The total number of prefetches to allow
+ * @param {Function} [options.timeoutFn] - Custom timeout function
  */
-export default function (options) {
+export function listen(options) {
```
addyosmani: Could you talk a little about the thought process of switching from a default to using `listen`? In practice, I think most users are likely to copy/paste the default code snippet we provide (and so I don't view this as a huge ergonomics change), but would be good to walk through reasoning :)

src/index.mjs
addyosmani: > The only use I can see is when a user slices up their page(s) into smaller options.el containers, saying that ContainerA is high priority and the others aren't. 

I haven't (to date) observed users heavily relying on granular `options.el` containers. I'm open to us considering this not-necessary and evaluating if we need to circle back during an alpha release. SG?

src/index.mjs
```diff
@@ -63,40 +43,87 @@ function isIgnored(node, filter) {
...
+    (options.el || document).querySelectorAll('a').forEach(link => {
+      // If the anchor matches a permitted origin
+      // ~> A `[]` or `true` means everything is allowed
+      if (!allowed.length || allowed.includes(link.hostname)) {
+        // If there are any filters, the link must not match any of them
+        isIgnored(link, ignores) || observer.observe(link);
+      }
+    });
+  }, {
+    timeout: options.timeout || 2e3
```
addyosmani: Smallest of nits: could we expand `2e3`? :) I know we're code-golfing but would like to keep the source a wee bit readable for folks who aren't sure why we're using 2000.

src/index.mjs
addyosmani: 👍

src/index.mjs
```diff
@@ -63,40 +43,87 @@ function isIgnored(node, filter) {
...
+
+
+/**
+* Prefetch a given URL with an optional preferred fetch priority
+* @param {String} url - the URL to fetch
+* @param {Boolean} [isPriority] - if is "high" priority
+* @param {Object} [conn] - navigator.connection (internal)
+* @return {Object} a Promise
+*/
+export function prefetch(url, isPriority, conn) {
```
addyosmani: 👍 to exposing `prefetch` like this.

I am excited about is us enabling Quicklink to be viewed as both a library with a heuristic for prefetching but also as just a generally good prefetcher. I believe hiding prefetch.mjs behind our abstraction likely led to users thinking you had to use our heuristic to take advantage of any of the prefetching behavior at all, which isn't the case.

src/index.mjs
lukeed: Maybe there's not as much consensus as I think there is, but I think have mixed imports vs default-only and/or named-only is the trickiest of the three. So, for me it basically comes down to convention/simplicity. It's certainly not impossible to have mixed exports.

However, for UMD and CJS users, it does mean you have to call "listen" function via `quicklink.default()`

src/index.mjs
lukeed: Are you suggesting we drop `options.el` too? Or are you saying you're also on the fence about `options.priority`'s relevance? Sorry 🙈

lukeed:
> I would be supportive of dropping urls if we modified the prefetcher to accept either a single URL string or an array of URLs (to preserve ergonomics and ease upgrading for anyone switching to this version). Wdyt?

I was envisioning the a `Promise.all()` from userland to achieve this. It doesn't really matter to me, but my thinking was that if they cared about the responses, their going to have to loop the return array anyway

lukeed:
Pushed up some changes, including a look at having `prefetch` accept multiple URLs.

Once we have the change list ironed out, I think I'll wrap this up with a set of test & doc updates.

This is updated build stats:

```
        914 B: quicklink.js.gz
        777 B: quicklink.js.br
        921 B: quicklink.mjs.gz
        789 B: quicklink.mjs.br
        992 B: quicklink.umd.js.gz
        842 B: quicklink.umd.js.br
```

src/index.mjs
addyosmani: No worries!

After digging in further with a few users, it looks like folks _do_ need the ability to limit observation to a set of elements (e.g just navigation UI or just DOM elements corresponding to article content but not nav). That said.. I'm definitely on the fence about `options.priority`.

When it was first implemented, I thought there was a lot of value in providing the ability to fetch next pages/next resources with a high priority if it was properly scheduled. However, in retrospect, I don't believe this is actually necessary. If we just supported rel=prefetch and fallbacks without this notion of `priority` that would still serve the 90% use-case and also provide a way to further trim down bundle size.

Wdyt?

addyosmani:
> Pushed up some changes, including a look at having prefetch accept multiple URLs.

Thank you, @lukeed!

src/index.mjs
lukeed: We'd only be saving 9B by not having this option inside `listen` method – size isn't my driver here. I'm just not sure it has benefit in an IO context.

But, if we're certainly keeping `options.el`, then there _is_ at least one use case for this.

```
listen({
  el: document.getElementById('nav'),
  priority: true
})

listen({
  el: document.getElementById('wrapper'),
  // regular priority
  throttle: 3
})
```

I'm fine keeping it as is – someone somewhere will want to do this

addyosmani: Keeping this as is sounds reasonable to me

addyosmani:
Catching up on the latest discussions, I believe we now have quorum? :) Thanks again. If there's anything design-wise you're waiting on further input for please feel free to reach out anytime.

I think moving on to tests and docs (whenever you have time) should feel unblocked.

lukeed:
Hey hey, here are final build stats

```
Build "quicklink" to dist:
        906 B: quicklink.js.gz
        763 B: quicklink.js.br
        910 B: quicklink.mjs.gz
        775 B: quicklink.mjs.br
        987 B: quicklink.umd.js.gz
        837 B: quicklink.umd.js.br
```

Not sure how to test throttle – not familiar enough with all puppeteer APIs. Follow up with docs in the morning~

addyosmani:
Thanks for working through these changes, @lukeed! Tests LGTM so far.

lukeed:
Hey, sorry for the delay on this one 🙈 

Good news is that I figured how to test the throttling~!
Updated the readme too.

Let me know if you need anything to be changed. Now that the bulk of the work is out of the way, anything else should be much quicker, even with my current forever-catchup schedule.

lukeed:
Hey @addyosmani, I reformatted the README in a way that reads easier _to my eyes_ – I felt like the  `listen()` docs were crammed, especially with the addition of the `prefetch()` text behind it.

They are here (https://github.com/lukeed/quicklink/pull/1 – [rendered](https://github.com/lukeed/quicklink/blob/28f372e989d4b2056f38d375bcfd93fc5e9ec248/README.md)). If you like them, I'll merge that PR which will update this guy.

addyosmani:
Rendered README LGTM. Thanks for reworking it, Luke! Just read through.

addyosmani:
W00. It's landed. Thanks, Luke!

giphy (43)

PR #122, opened 2019-09-24
Update docs with ad-related considerations

demianrenzulli:
Adding sub-section with considerations to take for sites monetizing through ads.

cc // @addyosmani


addyosmani:
LGTM. Thank you!

PR #123, opened 2019-09-27
feat(options): Add a `hrefFn` option to build the URL to prefetch.

ruben0626:
### Description
- adds `hrefFn` option - If this function is supplied, then it will be used to generate the URL to prefetch instead of using the `entry.href`.

### Why ?
I'm building a SPA which renders the links with `href` to a HTML page, however, there is an `onClick` listener for every link:
1. It fetches a Rest API sending the link `href`. The API returns a JSON.
2. The JSON is used to update the state of the SPA so it will update the UI to render the new page.

I need to prefetch the URL for the Rest API rather than the URL for the actual page.
```
<a href="https://www.webapage.com/article-1">Article 1</a>
```
 I want to prefetch https://rest-api.com/?url=https://www.webapage.com/article-1

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,11 @@ Default: None
 An optional error handler that will receive any errors from prefetched requests.<br>
 By default, these errors are silently ignored.
 
+#### options.hrefFn
+Type: `Function`<br>
+Default: None
+
+An optional function to generate the URL to prefetch. It receives an [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) as the argument.
 
 ### quicklink.prefetch(urls, isPriority)
 Returns: `Promise`
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -14,7 +14,7 @@
  * limitations under the License.
 **/
 import throttle from 'throttles';
-import { priority, supported } from './prefetch.mjs';
+import {priority, supported} from './prefetch.mjs';
 import requestIdleCallback from './request-idle-callback.mjs';
 
 // Cache of URLs we've prefetched
@@ -52,6 +52,9 @@ function isIgnored(node, filter) {
  * @param {Number} [options.limit] - The total number of prefetches to allow
  * @param {Function} [options.timeoutFn] - Custom timeout function
  * @param {Function} [options.onError] - Error handler for failed `prefetch` requests
+ * @param {Function} [options.hrefFn] - Function to use to build the URL to prefetch.
+ *                                             If it's not a valid function, then it will use the entry href.
+ * @return {Function}
  */
 export function listen(options) {
   if (!options) options = {};
@@ -64,6 +67,7 @@ export function listen(options) {
   const ignores = options.ignores || [];
 
   const timeoutFn = options.timeoutFn || requestIdleCallback;
+  const hrefFn = typeof options.hrefFn === 'function' && options.hrefFn;
 
   const observer = new IntersectionObserver(entries => {
     entries.forEach(entry => {
@@ -72,7 +76,7 @@ export function listen(options) {
         // Do not prefetch if will match/exceed limit
         if (toPrefetch.size < limit) {
           toAdd(() => {
-            prefetch(entry.href, options.priority).then(isDone).catch(err => {
+            prefetch(hrefFn ? hrefFn(entry) : entry.href, options.priority).then(isDone).catch(err => {
               isDone(); if (options.onError) options.onError(err);
             });
           });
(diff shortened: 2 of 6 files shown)


thearegee:
Any thoughts on getting this package merged?

taracila:
Hi @ruben0626, this is great feature!
By any chance, did you try to reduce network speed in the browser to 'Slow 3G' and see is you get any promise errors? Thank you

rockeynebhwani:
@addyosmani  - Did this get merge in Quicklink 2.0 or not? I don't see any mention of hrefFn option in documentation on https://getquick.link/ but this issue has been closed.

addyosmani:
@rockeynebhwani This hasn't been merged into Quicklink, as we punted on this particular solution in favor of recommending folks use our formal support for React + Quicklink covered in the README as part of the last release. That said, it's been brought to my attention that there are additional use-cases this PR would unlock that are not addressed by the other work done so I am re-opening it for consideration.

gilbertococchi:
Thanks Addy for reopening this FR, I am curious to here more feedback from other developers as well.

@thearegee would this PR make it viable to use QuickLinks to generate an API call for all your YNAP brands?

gilbertococchi:
Also found an interesting fork from @gibkigonzo that he seemed to expanded QuickLink to make it work on Vue StoreFront https://github.com/gibkigonzo/vsf-quicklink/tree/master/quicklink (discussed https://github.com/DivanteLtd/vue-storefront/issues/4161).

@gibkigonzo do you think this PR would be useful for your case as well?

gilbertococchi:
@thearegee Can you please let us know if you would find Quicklinks Callback helpful for your company sites?

Would this PR allow to easily integrate prefetch experiences?

gibkigonzo:
@gilbertococchi Finally I didn't use quicklink. It's great lib but in my case it was too aggressive. imo request to api is very heavy action so I didn't want to fire it for all visible links. So I used https://instant.page/ because it triggers event only for hovered links. Also I didn't need to fetch html so finally I end up with forked version of instant.page that allows to pass callback for each link https://github.com/gibkigonzo/instant-prefetch 
imo `hrefFn` can be useful for `quicklink`, for example you can use it for GTM events :shrug:

gilbertococchi:
Hi @gibkigonzo thanks for your feedback!

This is exactly the reason why I found your input really valuable!

I agree with you that Instant.page-like prefetching is less aggressive but I also noticed with several tests that is more impactful on Desktop browsing Experience Speed up while for Mobile connections prefetching on mouseover which is too close to touchend will not bring any benefit to Mobile navigations.
Did you manage to measure good results of mouseover prefetching on Mobile too?

I often suggest to better control Quicklink in a similar situation like the PWA-Studio or vue-storefront one by adjustin the "timeout" value to avoid sending too many API calls when users scroll but instead trigger them only when the user has finished scrolling for x seconds.

This PR is exactly about allowing a custom callback to prefetch an API Call instead of the HTML to use Quicklink, if merged do you think it would be interesting to test in Vue-storefront if usage is kept safe with "timeout" and other settings?

@addyosmani FYI
@thearegee FYI as resolving this PR conflicts will allows YNAP YOS to test Quicklink on these brands as well.

Gilberto

gibkigonzo:
I agree about mobile and this is also available in instant.page https://instant.page/intensity You can change `data-instant-intensity="viewport"` in body and then prefetch will be called by viewport change. imo at the end it always depend on layout and how many links we have visible. If we have one product on phone screen then viewport is best way, but if we have four products visible then maybe it's not best option. imo it should be configurable if we want to make api requests.

> This PR is exactly about allowing a custom callback to prefetch an API Call instead of the HTML to use Quicklink

I'm not sure, but in this pr we will still always prefetch html https://github.com/GoogleChromeLabs/quicklink/pull/123/files#diff-0b5856b7606f07e47d576a66791b7666ff3239ee4e354656cabefb7d5fecec88R132-R134
I think that it could be useful to replace html fetch by some custom action. In the end dev can still fetch html **and** some custom resources. 

> if merged do you think it would be interesting to test in Vue-storefront if usage is kept safe with "timeout" and other settings?

I've created data prefetch with vsf1 in mind and I will stop on extended version of instant.page. Maybe vsf core members will use it for vsf-next version (it uses Nuxt and is available on branch `next`). I think there may be a lot of cases for optimizations if we would have oportunity to fetch more then html. FYI @Fifciu

addyosmani:
Closed via https://github.com/GoogleChromeLabs/quicklink/pull/201

PR #129, opened 2019-10-20
Check for `IntersectionObserverEntry.prototype.isIntersecting`

dieulot:
quicklink is using `isIntersecting` but not all browsers that implement IntersectionObserver implement isIntersecting. The culprits are Edge 15 and Chrome 51–57. UC Browser is based on Chromium 57 so this is useful even for a “modern” browser.

Note that I didn’t test this (but it’s the pattern I use for instant.page).


XhmikosR:
I'm not sure if this is needed since the instructions clearly mention that one should use a polyfill?

addyosmani (approved):
I would be comfortable with us landing this change, while being mindful that the browser landscape has continued to evolve since the initial PR and the audience that may benefit from the change could be much smaller now.

dieulot:
[UC Turbo](https://play.google.com/store/apps/details?id=com.ucturbo) and/or [UC Mini](https://play.google.com/store/apps/details?id=com.uc.browser.en) were still on Chromium 57 last time I checked (I can’t be bothered to re-check). They have 100M+ and 50M+ downloads, versus UC Browser which has 1B+. That’s outside China, where they’re apparently the most popular. UC Browser’s main Chinese version (which is different from the worldwide version) recently jumped to Chromium 100 according to user agent strings online.

PR #148, opened 2019-12-06
Over-prefetching section for home page

demianrenzulli:
Section covering techniques to avoid over-prefetching with Quicklink.

### DIFF
--- a/site/src/_includes/components/over-prefetching.njk
+++ b/site/src/_includes/components/over-prefetching.njk
@@ -1,0 +1,46 @@
+{% extends "layouts/normal-section-wrapper.njk" %}
+{% block section %}
+{% sectionTitle "Concerned about over-prefetching? We've got you covered" %}
+{{ "By default `quicklink` observes all in-viewport links in `document.body`. There are different ways of telling `quicklink` to limit the number of links to prefetch." | markdown | safe }}
+
+{{ "The most common approach is passing passing different `options` to configure prefetching when calling `quicklink.listen()`:" | markdown | safe }}
+
+{{ "* Indicating a specific DOM element to observe, with the `options.el` parameter:" | markdown | safe }}
+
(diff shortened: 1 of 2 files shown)


site/src/_includes/components/over-prefetching.njk
```diff
@@ -0,0 +1,46 @@
...
+{{ "By default `quicklink` observes all in-viewport links in `document.body`. There are different ways of telling `quicklink` to limit the number of links to prefetch." | markdown | safe }}
+
+{{ "The most common approach is passing passing different `options` to configure prefetching when calling `quicklink.listen()`:" | markdown | safe }}
+
+{{ "* Indicating a specific DOM element to observe, with the `options.el` parameter:" | markdown | safe }}
+
+{% markdownConvert %}
+```js
+quicklink.listen({
+  el: document.getElementById('carousel')
```
addyosmani: Wdyt about us using an example like `document.getElementById('article')` or `document.getElementById('content')`?

demianrenzulli: Changed it to 'content', thank you!

addyosmani:
I think these changes strike the right balance of conveying how our latest features for limiting and controlling throttling help avoid the over-prefetching problem well. 

I like that we started off with sharing how to limit prefetching to specific elements. Something I may try doing if there's time is creating a nice SVG that shows how this works.

But overall I think this will help. Going ahead and landing the change 👍

demianrenzulli:
Great! Thanks @addyosmani. I tested both option parameters on the demo site (limit and throttling), and could visualize clearly in the network tab how in one case requests are limited to a certain number and, in the other, how they are sent sequentially instead of simultaneously.

With that said, if at some point think we would need a demo to complement this explanation, it can be easily done as well. 👍

PR #156, opened 2020-01-07
(feature) Adds prefetchChunks support

anton-karlovskiy:
@addyosmani @demianrenzulli 

`prefetchChunks` option is added for prefetching chunks on route based code-splitting with React.

`prefetchChunks` method is introduced specifically for the purpose of fetching chunk URLs (e.g. /static/js/about.f0141832.chunk.js) out of route URLs (e.g. /about) by mapping each route (/about etc) back to the Webpack chunk it corresponds to.

So this method consists of two parts, where:
* one part is for mapping with the help of [route-manifest](https://github.com/lukeed/route-manifest) package
* the other one for fetching mapped chunk URLs. Fetching is performed by `quicklink`'s `prefetch` method internally

```js
const prefetchChunks = (entry, prefetchHandler) => {
  try {
    const chunkEntry = rmanifest(window._rmanifest_, entry.pathname);
    const chunkURLs = chunkEntry.files.map(file => file.href);
    if (chunkURLs.length) {
      prefetchHandler(chunkURLs);
      return;
    }
  } catch (error) {
    console.log('[prefetchChunks] error => ', error);
  }

  // also prefetch regular links in-viewport
  prefetchHandler(entry.href);
};
```

We pass this `prefetchChunks` method as `quicklink` option so if this option is set then `quicklink` fetches chunks URLs instead of route URLs (of course regular links too as mentioned [here](https://github.com/addyosmani/react-quicklink-demo/pull/2#issuecomment-572211523) and @demianrenzulli [reply](https://github.com/addyosmani/react-quicklink-demo/pull/2#issuecomment-572232377) to it)

The `prefetchHandler` is passed as argument to `prefetchChunks` inside `quicklink` since fetching is ultimately done by `prefetch` method. It's just `prefetch` method with some options and error handling just to avoid code duplication.

The quicklink is normally designed to fetch regular links (hrefs) for routes in-viewport but once we pass `prefetchChunks` as option, it can fetch chunks by mapping hrefs back to chunks. If not, the workflow is the same as before.

```js
const prefetchHandler = urls => {
  prefetch(urls, options.priority).then(isDone).catch(err => {
    isDone(); if (options.onError) options.onError(err);
  });
};
```

This option is useful for fetching chunks in Single Page Applications like React SPA app.


addyosmani:
Thank you for your continued work on this PR, Anton. 

To provide some public notes on the changes here: we have spent the last month exploring React and single-page application support for Quicklink in a few private repos. This specifically involved solving this problem:

- A user had a single-page app that used a router (e.g React Router)
- Links to SPA routes Quicklink would prefetch would not fetch the actual JS needed to render a route; so prefetching /about, /contact etc were effectively not useful, even if links to them were in view. This is a problem almost all prefetching libraries don't help with.
- We needed a way to map each route (/about etc) back to the Webpack chunk it corresponded to. We also needed this to work with long-term hashing (e.g about.kjhas0dh0a.js)

@lukeed was kind enough to make a go at an initial solution here and we've been building on it. Some of the edge-cases we ran into were duplicate prefetch requests being made, in-viewport links not always being fetched etc, but have been trying to smooth these over. Software is hard 😄 

Where we are at is we've defined a solution to this problem that worked for React apps. If a user had set this up correctly, any in-viewport links would have their final JS prefetched correctly. Early attempts at introducing this solution to Quicklink via this PR were too highly coupled to React and bloated Quicklink's size e.g one of our dependencies is @lukeed's awesome https://github.com/lukeed/route-manifest, but took our overall bundle size in this PR to over 1.5KB. I didn't feel this was worthwhile as a cost to all Quicklink users.

The current PR has been trimmed back to just introduce a new `prefetchChunks()` method, which does increase our bundle size but by far less.

Capture

Once this lands, users will use `prefetchChunks` together with rmanifest and a new Quicklink HOC (`withQuicklink`) for React apps to get everything working the way they would expect:

```js
import React, { useEffect, useRef, useState } from 'react';
import { listen } from 'quicklink';

const useIntersect = ({ root = null, rootMargin, threshold = 0 }) => {
  const [entry, updateEntry] = useState({});
  const [node, setNode] = useState(null);
  const observer = useRef(null);

  useEffect(
    () => {
      if (observer.current) observer.current.disconnect();
      observer.current = new window.IntersectionObserver(
        ([entry]) => updateEntry(entry),
        {
          root,
          rootMargin,
          threshold
        }
      );

      const { current: currentObserver } = observer;
      if (node) currentObserver.observe(node);

      return () => currentObserver.disconnect();
    },
    [node, root, rootMargin, threshold]
  );

  return [setNode, entry];
};

const withQuicklink = (Component, options) => {
  return props => {
		const [ref, entry] = useIntersect({root: document.body.parentElement});
    const intersectionRatio = entry.intersectionRatio;
    
		useEffect(() => {
			if (intersectionRatio > 0) {
        listen(options);
			}
		}, [intersectionRatio]);
		
		return (
			<div ref={ref}>
        <Component {...props} />
			</div>
		);
	};
};

export {
  withQuicklink
};
```

I believe https://glitch.com/~anton-karlovskiy-intersection-observer-react-quicklink-demo has the latest demo of these pieces working together. @lukeed in case interested, on the evolving discussions that got us here happened over on [this PR](https://github.com/addyosmani/react-quicklink-demo/pull/).

src/index.mjs
```diff
@@ -65,16 +65,28 @@ export function listen(options) {
 
   const timeoutFn = options.timeoutFn || requestIdleCallback;
 
+  // ray test touch <
```
addyosmani: Could we remove the `ray test touch` lines from source now that we're closer to a solution we think is workable?

test/quicklink.spec.js
```diff
@@ -1,5 +1,8 @@
 describe('quicklink tests', function () {
   const server = `http://127.0.0.1:8080/test`;
+  // ray test touch <
+  const host = `http://127.0.0.1:8080`;
```
addyosmani: It would probably make more sense to change these lines as follows:

```js
const host = `http://127.0.0.1:8080`;
const server = `${host}/test`;
```

src/index.mjs
```diff
@@ -65,16 +65,28 @@ export function listen(options) {
...
+  const prefetchChunks = options.prefetchChunks;
```
addyosmani: We are reading `prefetchChunks` from the `options`,  but have not written JSDoc comments for it in the comments higher up (the same way other options are documented). This should probably be changed.

src/index.mjs
```diff
@@ -65,16 +65,28 @@ export function listen(options) {
...
+  // ray test touch >
+
+  // ray test touch <
+  const prefetchHandler = urls => {
+    prefetch(urls, options.priority).then(isDone).catch(err => {
+      isDone(); if (options.onError) options.onError(err);
+    });
+  };
```
addyosmani: We should include a PR comment to explain why `prefetchHandler` is needed. Could you add a comment reply below here?  For other contributors, they know `prefetch()` already supports passing an array of URLs but may not be certain why introducing this new `prefetchChunks` option is needed.

src/index.mjs
```diff
@@ -65,16 +65,28 @@ export function listen(options) {
...
       if (entry.isIntersecting) {
         observer.unobserve(entry = entry.target);
         // Do not prefetch if will match/exceed limit
         if (toPrefetch.size < limit) {
           toAdd(() => {
-            prefetch(entry.href, options.priority).then(isDone).catch(err => {
-              isDone(); if (options.onError) options.onError(err);
-            });
+            // ray test touch <
+            prefetchChunks ? prefetchChunks(entry, prefetchHandler) : prefetchHandler(entry.href);
```
addyosmani: I think once we have explained why the `prefetchChunks` option is needed in this PR, it would be good to also write out some docs additions to the project README so that we have API docs for this change too.

src/index.mjs
anton-karlovskiy: Yes for sure, I think it's time.

test/quicklink.spec.js
anton-karlovskiy: Thank you for teaching.
Yes, it makes way more sense.

For now, I have not updated the Unit Test accordingly.
As you think the latest updates are close enough, I'm going to update the Unit Test.

src/index.mjs
anton-karlovskiy: So we also should document prefetchChunks option in README as other options are documented like [here](https://github.com/GoogleChromeLabs/quicklink#api).

src/index.mjs
anton-karlovskiy: `prefetchChunks` method is introduced specifically for the purpose of fetching chunk URLs (e.g. /static/js/about.f0141832.chunk.js) not route URLs (e.g. /about) by mapping each route (/about etc) back to the Webpack chunk it corresponded to.

So this method consists of two parts, where:
* one part is for mapping
* the other one for fetching mapped chunk URLs. Fetching is done by `quicklink`'s `prefetch` method.
```
const prefetchChunks = (entry, prefetchHandler) => {
  try {
    const chunkEntry = rmanifest(window._rmanifest_, entry.pathname);
    const chunkURLs = chunkEntry.files.map(file => file.href);
    if (chunkURLs.length) {
      prefetchHandler(chunkURLs);
      return;
    }
  } catch (error) {
    console.log('[prefetchChunks] error => ', error);
  }

  // also prefetch regular links in-viewport
  prefetchHandler(entry.href);
};
```

We pass this `prefetchChunks` method as `quicklink` option so if this option is set then `quicklink` fetches chunks URLs instead of route URLs (of course regular links as mentioned [here](https://github.com/addyosmani/react-quicklink-demo/pull/2#issuecomment-572211523) and @demianrenzulli [reply](https://github.com/addyosmani/react-quicklink-demo/pull/2#issuecomment-572232377) to it)

This `prefetchHandler` is passed as argument to `prefetchChunks` since fetching is ultimately done by `prefetch` method. It's just `prefetch` method with some options and error handling just to avoid code duplication.

The quicklink is designed to fetch regular links(hrefs of anchor tags) in-viewport but when if we pass `prefetchChunks` as option, it fetches chunks by it. If not, the workflow is the same as before.
```
prefetchChunks ? prefetchChunks(entry, prefetchHandler) : prefetchHandler(entry.href);
```

This option is useful for fetching chunks in Single Page Applications like React SPA app.

lukeed: Hmm.. I still don't understand why the separate `prefetchHandler` is needed, sorry 🙈 Can you explain why [this approach](https://github.com/lukeed/route-manifest#usage) doesn't work?

src/index.mjs
anton-karlovskiy: I added some comment above regarding this.
Once it's confirmed, I'm going to update README and PR comment accordingly.

src/index.mjs
anton-karlovskiy: Actually it's not needed to use the separate `prefetchHandler`. We can just use that approach.

I just tried to replicate [this](https://github.com/GoogleChromeLabs/quicklink/blob/836f170df4007c05c24b1300e0faf04fb823168b/src/index.mjs#L75)

Shall we update it?

test/quicklink.spec.js
anton-karlovskiy: Done

src/index.mjs
anton-karlovskiy: Done

src/index.mjs
anton-karlovskiy: Done

src/index.mjs
anton-karlovskiy: Done

src/index.mjs
addyosmani: @lukeed if you have time to do a second round of reviews, that would be super appreciated. Going to take another look at the latest commits now too.

addyosmani: @anton-karlovskiy Looking at https://github.com/lukeed/route-manifest#usage, I feel we may still be missing why we can't achieve the same behavior as `prefetchChunks()` using Quicklink's current `prefetch()` utility. 

Stepping back...

Step 4 in https://github.com/anton-karlovskiy/intersection-observer-react-quicklink-demo#how-to-setup asks developers to use a React HOC in order to get Quicklink working for routes. 

What if `prefetchChunks` was just a part of the HOC code and re-used `prefetch()`?

```js
const prefetchChunks = (entry) => {
  try {
    const chunkEntry = rmanifest(window.__rmanifest, entry.pathname);
    const chunkURLs = chunkEntry.files.map(file => file.href);
    if (chunkURLs.length) {
      prefetch(chunkURLs);
      return;
    }
  } catch (error) {
    console.log(error);
  }
  // also prefetch regular links in-viewport
  prefetch(entry.href);
};
```

That way we would just package https://glitch.com/edit/#!/anton-karlovskiy-intersection-observer-react-quicklink-demo?path=src/lib/quicklink.js:49:0, but could get everything working without expanding Quicklink's API surface.

Would that work? (i.e what we PR in to this repo is the HOC but not land this PR). Or am I missing something? :)

anton-karlovskiy: @addyosmani cc @lukeed 

Yes, I agree with you. That should be much better option.
I just tried to replicate the usage of options. But no need, come to think of it.

anton-karlovskiy: @addyosmani cc @lukeed 

Sorry :)
Let me think of it carefully once more and update you.

lukeed: Yeah, I agree. I really don't think there's anything _new_ to be added to `quicklink` here – at least when talking about `prefetch`'s ability to work in its current form. 

Also I'm not even sure that the HoC is needed...but that may just be me.

The "problem" is that the route manifest can have varying shapes. The `route-manifest` package can only traverse the manifest and return a matching `{ files, headers }` object, but _what_ those `files` look like is entirely up to the developer.  By default, `webpack-route-manifest` produces entries that can fulfill `<link rel=preload>` and `Link` HTTP headers – but many projects may only ever care about the URL, for example, and so their manifest entries may skip the object & return URLs only.

If we were to include a HoC, we'd have to cover both use cases (no problem) ***or*** define an interface that requires a `callback` argument, which is used to extract the URL from each of the `files` returned.

This also kinda assumes that `route-manifest` be included as a quicklink dependency. Otherwise including a utility to filter down to URLs doesn't really make sense? Maybe I'm missing something, it's been a while.

anton-karlovskiy: After I've discussed with @addyosmani, we made a decision to explore the HOC approach and try to address the 90% of use-cases with it. But we would like to review with @luke if he thinks it makes sense once we have something working.

lukeed: Sounds good! Let me know if you need anything 👍

anton-karlovskiy: Thank you @lukeed.

I'm working on it and update you and @addyosmani.

addyosmani:
To share an update here:

We are most likely going to move ahead with the introduction of a higher order component (HOC) for React users with minimal changes to Quicklink's API surface to support this. 

Work on this has taken longer than we originally anticipated, but we are trying to get it right with minimal steps required by the developer.

addyosmani:
Closed via https://github.com/GoogleChromeLabs/quicklink/pull/171

PR #171, opened 2020-04-13
Introduce prefetch chunks build

addyosmani:
This is a follow-up to https://github.com/GoogleChromeLabs/quicklink/pull/156 which adds support for prefetch chunks as a separate build. The goal of this build is to give developers who are building single-page applications to logic required to do so, without requiring all users to pay the extra byte-costs.


anton-karlovskiy:
@addyosmani cc @demianrenzulli 

LGTM!

demianrenzulli:
Just went through the code and couldn't find any issues there, so it LGTM too 👍

addyosmani:
Thank you, @anton-karlovskiy and @demianrenzulli!

PR #172, opened 2020-04-19
[core] Adds withQuicklink HOC

addyosmani:
Usage after installing the package:

```js
import { withQuicklink } from 'quicklink/react/hoc.mjs';

<Suspense fallback={<div>Loading...</div>}>
  <Route path="/" exact component={withQuicklink(Home, options)} />
  <Route path="/blog" exact component={withQuicklink(Blog, options)} />
  <Route path="/blog/:title" component={withQuicklink(Article, options)} />
  <Route path="/about" exact component={withQuicklink(About, options)} />
</Suspense>
```

cc @anton-karlovskiy @demianrenzulli fyi

@demianrenzulli Do you think we could try using this build locally with the latest demo app and see if it works as you would expect?


src/react-chunks.mjs
```diff
@@ -0,0 +1,86 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { useEffect, useRef, useState } from 'react';
+import rmanifest from 'route-manifest';
+import { listen } from './quicklink';
```
demianrenzulli, line 19: Just to check: we don't need to specify here other version of the library rather than ./quicklink?
Asking in case the spa version ends up being installed on a different package?

// @addyosmani @anton-karlovskiy

addyosmani: My current thinking is that everything inside the dist directory will get published to npm. This should allow this file to just reference the local (chunks) build of Quicklink.

I think that this would be more straightforward than introducing a secondary package that we publish?

anton-karlovskiy:
@addyosmani 

I debugged the failed tests.
At line 20 of  `quicklink\test\test-prefetch-chunks.html`:

If we adjust the path:
from
```
<script src="../dist/chunks/quicklink.umd.js"></script>
```
to
```
<script src="../dist/react/quicklink.umd.js"></script>
```
Then we can pass the tests.

anton-karlovskiy:
@addyosmani cc @demianrenzulli 

I think we can uninstall `babel-preset-env` as we installed upgraded version of `@babel/preset-env`.

addyosmani:
I think it's just about time we landed this :)

A big thank you to everyone that has been working with us to research and implement this change!

PR #178, opened 2020-04-26
[infra] Add site updates for 2.0.0

addyosmani:
In-flight site updates for 2.0.0.


demianrenzulli:
@addyosmani now we have approval from the company, changes LGTM 👍

addyosmani:
@demianrenzulli fantastic. Yay.

Any other site-related changes (other than bumping the version number + tag) that we're wanting to land for this release? (e.g is the React demos related work you and Anton are driving planned for the README or README + site?)

demianrenzulli:
Good catch! I'd like to edit the demo section of the site to clearly separate between MPAs and SPAs and the latest ones.

Do you think the current ones deployed in Glitch can be used for that? Or do you think we'll end up uploading it somewhere else? If not, I can work on these changes for the site as well.

demianrenzulli:
Perhaps it's a better to just move forward with this PR, and then I can take care of updating the site once we have decided where to host these demos? That way I don't block these changes.

addyosmani:
SG.

Issue #182, opened 2020-05-16
Observe more than 1 DOM element


cajotafer:
The site I want to try to apply this is a news site. There is a main DOM element I want to observe but there are also other containers I would like to observe at the same time and I see right now I can only set 1 DOM element.

I would like to add not just 1 DOM element to be observed as an option, but more than 1 element, maybe as an Array.

I don't know if the "simple" solution will have a performance issue.

cajotafer:
Uhm, in this case I image two simple approaches in the listen function.

1. Using the same `options.el` to receive the node or an array of nodes. In case an array is comming, merge the content of the array into a new array and use the `querySelectorAll()` here.
2. The same than n.1 but using something like `options.elements` instead of `options.el` for semantic purposes.

addyosmani:
This a reasonable FR. We can expand `el` to accept an array of nodes and iterate through them as necessary (closer to option 1 above)

PR #183, opened 2020-05-17
[core] Observe more than 1 DOM element

cajotafer:
You are doing something great guys and I want to contribute.
I am newbie contributing so I appreciate you understand. If there is some mistake or improvements to do I will be more than happy to receive your feedback.

This PR is related to the FR in the issue #182 . I modified the `index.mjs` to accept an HTMLelement or an array of HTMLelements and iterate through them as necessary to observe the links. 
I also tried to write the test and update the `README`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,10 +114,10 @@ Returns: `Function`
 A "reset" function is returned, which will empty the active `IntersectionObserver` and the cache of URLs that have already been prefetched. This can be used between page navigations and/or when significant DOM changes have occurred.
 
 #### options.el
-Type: `HTMLElement`<br>
-Default: `document.body`
+Type: `HTMLElement|Array<HTMLElement>`<br>
+Default: `document`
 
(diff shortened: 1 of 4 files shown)


gilbertococchi:
Great change, I like it and I think it would make Quicklink even more useful for developers.

rangerz:
Consider not changing the codebase to use `options.ignores` for the alternative.
Provide `Function` to filter which Elements need to be prefetched or not.
For example in your [test html](https://github.com/GoogleChromeLabs/quicklink/blob/49912c9f0f4eedcae854d9df664a44ae51b4f9a9/test/test-custom-dom-source-multiple.html)
```Javascript
quicklink.listen({
    ignores: [
        (uri, elem) => !(
            document.getElementById('stuff1').contains(elem) ||
            document.getElementById('stuff2').contains(elem)
        )
    ],
});
```

PR #188, opened 2020-05-21
Fix: Cannot read property 'then' of undefined

leepowelldev:
This PR fixes a `Cannot read property 'then' of undefined` error that gets thrown when this condition is met:
```
if (conn = navigator.connection) {
  // Don't prefetch if using 2G or if Save-Data is enabled.
  if (conn.saveData || /2g/.test(conn.effectiveType)) return;
}
```
`prefetch` is meant to return a Promise, however in this instance it returns `undefined`, meaning that this throws:
```
prefetch(entry.href, options.priority).then(isDone).catch(err => {
  isDone(); if (options.onError) options.onError(err);
});
```
I tried to add tests for this, but I could not figure out a way of mocking either `saveData` or a 2g connection - I tried multiple approaches, with none working. If anyone has any ideas then I'm happy to add the tests to cover this.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": false,
+    "source.organizeImports": false
+  },
+  "editor.formatOnSave": true
+}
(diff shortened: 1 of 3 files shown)


leepowelldev:
@googlebot I signed it!

leepowelldev:
@googlebot I signed it!

addyosmani (approved):
Thanks, Lee!

PR #207, opened 2020-11-19
[chore] Fix local tests

addyosmani:
Demian and I were having difficulties getting our tests to run locally, while they were completely fine in CI from what I can ascertain. Made some local changes to bump the timeouts and change the port serving number (...surprised this wasn't an issue in CI) which makes things pass.

As long as CI continues to not complain with these changes in, I'm probably fine with landing them.


addyosmani:
Closing per discussion with Demian.

Issue #208, opened 2020-11-20
When `throttle` is set, cancel pending network requests of items that exited the viewport?


verlok:
**Problem**
When `throttle` is set to a number (say `4`) and users scroll down fast, all the links that entered and exited the viewport sit in a queue and they are fetched progressively, `4` at a time. This implies that if users then click on a link at the bottom of the page, the prefetch for the link they clicked might still not arrived (because QuickLink is still prefetching links of the upper part of the page), making QuickLink's efforts useless.

**Solution we'd like**
We'd like QuickLink to avoid fetching links that already exited the viewport, when the `throttle` is set. This would give priority to the links that are currently in viewport.

**Additional context**
We're trying to lower the number of network requests that hit our servers, and we thought that using `throttle` would be useful. Turns out that yes, it preserves our servers to be bombed, but it becomes useless for the users who scroll down fast.

verlok:
Does this make sense to you @addyosmani? I'm keen to work at a pull request if it does.

verlok:
Closing this, we've found a better solution to lower the number of network requests that hit our servers. I'll open a new issue to explain.

PR #212, opened 2020-12-15
Added `delay` option to solve #208

verlok:
Added and tested the following option

#### options.delay
Type: `Number`<br>
Default: `0`

The _amount of time_ each link needs to stay inside the viewport before being prefetched, in milliseconds.


verlok:
I think I might have committed with the wrong email address. Now how can I fix this?

verlok:
Will do again using a single commit

PR #213, opened 2020-12-15
Added `delay` option to reduce impact on CDNs and servers

verlok:
Added the `delay` option, which is useful to prevent QuickLink from prefetching too much links and have a heavy impact on CDN and servers.

> Type: `Number`
> Default: `0`
> 
> The _amount of time_ each link needs to stay inside the viewport before being prefetched, in milliseconds.

Unlike `throttle` option which prefetches the links in order of appearance, this option makes sure the links _currently inside the viewport_ are prefetched.

So for example, having 40 links in the page and users scrolling down fast to the 40th...

- with `throttle: 4`, QuickLink prefetches the first 4 links, then the 5th, 6th, 7th ... and finally the 40th
- with `delay: 500`, QuickLink prefetches only the first ones that are in viewport, then if users scroll fast enough, the 39th, the 40th, etc. because the previous 38 links didn't stay in-viewport for 500 ms

Note that the `throttle` and `delay` options can still be used together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ Returns: `Function`
 
 A "reset" function is returned, which will empty the active `IntersectionObserver` and the cache of URLs that have already been prefetched. This can be used between page navigations and/or when significant DOM changes have occurred.
 
+#### options.delay
+Type: `Number`<br>
+Default: `0`
+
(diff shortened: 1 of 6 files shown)


verlok:
The alternative version was chosen and merged, this is not useful anymore.

PR #214, opened 2020-12-16
Added `threshold` option to reduce impact on CDNs and servers

verlok:
Added `threshold` option to allow users select the % of link areas that entered the viewport before prefetching, reduce the number of links to prefetch at a time, and ultimately reduce the impact on CDNs and servers.

> Type: `Number`
> Default: `0`
> 
> The _area percentage_ of each link that must have entered the viewport to be fetched, in its decimal form (e.g. 0.25 = 25%).

So for example, having 4 links in the page, 2 of which completely in-viewport and other 2 that are barely in-viewport (for less than 25% of their area), setting `threshold: 0.25` QuickLink will prefetch only the first and the second link, which are fully in-viewport. Then it will prefetch the 3rd and the 4th link when users scroll down for > 25% of the next link areas.


verlok:
Hey @addyosmani, how about this PR? 
We are already using this option in a custom version of Quicklink but we'd like to see it in the official repo :)

verlok:
Hey @addyosmani, me again.

I created a small demo to find the most effective quicklink settings.

It's here:
https://www.andreaverlicchi.eu/quicklink-optimal-options/

Changing the values in the demo I realized that we can very well set the `threshold` to 0.9.

Long story short, ideal settings in my opinion are:

```js
{
   throttle: 4,
   threshold: 0.9,
   delay: 500
}
```

Which for us means that in order to fetch a PDP, its link must be in viewport for 90% of its extension, stay in this condition for 500ms, and no more than 4 other prefetches must be running.

I hope this convince you to merge this PR!
After all, it's a single line of code: passing the `threshold` option to the `IntersectionObserver` options.

addyosmani:
Hey @verlok. If we can resolve the merge conflicts this should be good to merge!

verlok:
Sure thing @addyosmani. I've just done that.

addyosmani (approved):
LGTM. Thanks for this contribution!

PR #217, opened 2020-12-18
Added `delay` option to reduce impact on CDNs and servers [alternative without data-attributes]

verlok:
📒 _This is an alternative to #213 without using data-attributes_

Added the `delay` option, which is useful to prevent QuickLink from prefetching too much links and have a heavy impact on CDN and servers.

> Type: `Number`
> Default: `0`
> 
> The _amount of time_ each link needs to stay inside the viewport before being prefetched, in milliseconds.

Unlike `throttle` option which prefetches the links in order of appearance, this option makes sure the links _currently inside the viewport_ are prefetched.

So for example, having 40 links in the page and users scrolling down fast to the 40th...

- with `throttle: 4`, QuickLink prefetches the first 4 links, then the 5th, 6th, 7th ... and finally the 40th
- with `delay: 500`, QuickLink prefetches only the first ones that are in viewport, then if users scroll fast enough, the 39th, the 40th, etc. because the previous 38 links didn't stay in-viewport for 500 ms

Note that the `throttle` and `delay` options can still be used together.

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -50,6 +50,7 @@ function isIgnored(node, filter) {
  * @param {Number} [options.timeout] - Timeout after which prefetching will occur
  * @param {Number} [options.throttle] - The concurrency limit for prefetching
  * @param {Number} [options.limit] - The total number of prefetches to allow
+ * @param {Number} [options.delay] - Time each link needs to stay inside viewport before prefetching (milliseconds)
  * @param {Function} [options.timeoutFn] - Custom timeout function
  * @param {Function} [options.onError] - Error handler for failed `prefetch` requests
  * @param {Function} [options.hrefFn] - Function to use to build the URL to prefetch.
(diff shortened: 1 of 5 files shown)


src/index.mjs
```diff
@@ -65,22 +66,49 @@ export function listen(options) {
...
-            prefetch(hrefFn ? hrefFn(entry) : entry.href, options.priority).then(isDone).catch(err => {
-              isDone(); if (options.onError) options.onError(err);
+        entry = entry.target;
+        // Adding href to array of hrefsInViewport
+        hrefsInViewport.push(entry.href);
+
+        // Setting timeout
+        setTimeoutIfDelay(() => {
+          const found = hrefsInViewport.indexOf(entry.href) > -1;
+          if (!found) return;
```
addyosmani, line 94: If `found` is not being reused, it may be worth modifying this inline to if (! < the condition>) return; rather than defining a constant.

addyosmani (changes requested):
Overall LGTM with only one nit. I would recommend also adding in a docs item to https://github.com/GoogleChromeLabs/quicklink/blob/master/site/src/api.njk as part of this PR to enable it to land.

src/index.mjs
verlok: Hey, thanks for the comment. Our intention was to make the code more readable, but it's your repo so we will change it.

verlok:
Hey @addyosmani, I guess you would release this in a new version of quicklink, e.g. 2.1.0, am I right?
That would be a minor change and that shouldn't bring breaking changes to the way quicklink works.

But for the next major version, say 3.0.0, what do you think to set the default value of the `delay` option to `500`?
I'm saying this because, from our experiments, the optimal values are the following.

```js
{
  delay: 500,
  threshold: 0.25
}
```

addyosmani:
> But for the next major version, say 3.0.0, what do you think to set the default value of the delay option to 500?

I would certainly be interested in us setting a better default value based on data driving config we've seen be successful in the wild :)

addyosmani:
> Hey @addyosmani, I guess you would release this in a new version of quicklink, e.g. 2.1.0, am I right?

Correct. We would be releasing this in a new non-major version

addyosmani (approved):

PR #258, opened 2022-04-26
Add support for same-site prerendering with Speculation Rules API

### LINKED ISSUE #253: Add same-site prerendering with speculation rule
Add support to [same-site prerendering](https://github.com/WICG/nav-speculation/blob/main/same-origin-explainer.md) using speculation rule.

hadyan:
- As a new "prerender" option for quicklink.listen to prerender link(s) in the viewport with fallback to prefetching on unsupported browsers.
- Create a new API quicklink.prerender to define static url(s) to prerender.

Closes #253


README.md
```diff
@@ -17,11 +17,11 @@ Quicklink attempts to make navigations to subsequent pages load faster. It:
 * **Detects links within the viewport** (using [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API))
 * **Waits until the browser is idle** (using [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback))
 * **Checks if the user isn't on a slow connection** (using `navigator.connection.effectiveType`) or has data-saver enabled (using `navigator.connection.saveData`)
-* **Prefetches URLs to the links** (using [`<link rel=prefetch>`](https://www.w3.org/TR/resource-hints/#prefetch) or XHR). Provides some control over the request priority (can switch to `fetch()` if supported).
+* **Prefetches** (using [`<link rel=prefetch>`](https://www.w3.org/TR/resource-hints/#prefetch) or XHR) or **prerenders** (using [Speculative Rules API](https://github.com/WICG/nav-speculation/blob/main/triggers.md))  URLs to the links. Provides some control over the request priority (can switch to `fetch()` if supported).
```
domenic, line 20:
```suggestion
* **Prefetches** (using [`<link rel=prefetch>`](https://www.w3.org/TR/resource-hints/#prefetch) or XHR) or **prerenders** (using [Speculation Rules API](https://github.com/WICG/nav-speculation/blob/main/triggers.md))  URLs to the links. Provides some control over the request priority (can switch to `fetch()` if supported).
```

I noticed this typo a few other times too...

README.md
```diff
@@ -380,6 +415,7 @@ Certain features have layered support:
 
 * The [Network Information API](https://wicg.github.io/netinfo/), which is used to check if the user has a slow effective connection type (via `navigator.connection.effectiveType`) is only available in [Chrome 61+ and Opera 57+](https://caniuse.com/#feat=netinfo)
 * If opting for `{priority: true}` and the [Fetch API](https://fetch.spec.whatwg.org/) isn't available, XHR will be used instead.
+* Prerendering using Speculative Rules API can be experimented on Chrome >= 94 on any platform by enabling the experimental flag `chrome://flags/#enable-prerender2` for local development. Live testing is available for Chrome for Android by signing up for the [origin trial](https://developer.chrome.com/origintrials/#/view_trial/1325184190353768449) and adding the [origin trial meta tag](https://github.com/WICG/nav-speculation/blob/main/same-origin-chrome-origin-trial.md#origin-trial-registration).
```
domenic, line 418: The origin trial is no longer necessary on Chrome for Android >= 103.

src/index.mjs
```diff
@@ -175,3 +222,45 @@ export function prefetch(url, isPriority, conn) {
...
+
+/**
+* Prerender a given URL
+* @param {String} url - the URL to fetch
+* @param {Object} [conn] - navigator.connection (internal)
+* @return {Object} a Promise
+*/
+export function prerender(urls, conn){
+  checkConnection(conn = navigator.connection).catch(err => {
+    return Promise.reject(new Error('Cannot prerender, '+err.message));
```
domenic, line 234: Nothing is catching this promise rejection (since it is inside a catch handler).

src/prerender.mjs
```diff
@@ -0,0 +1,60 @@
...
+  return true;
+}
+
+/**
+ * Add a given set of urls to the speculation rules
+ * @param {Set} toPrerender - the URLs to add to speculation rules
+ * @return {Object} a promise
+ */
+export function addSpeculationRules(urlsToPrerender){
+  return new Promise( (resolve, reject, specScript) => {
```
domenic, line 37: The promise constructor executor function only accepts two arguments, `(resolve, reject)`. You should probably declare `specScript` as a local variable.

README.md
addyosmani: 👍🏻 to Domenic's feedback here. Let's adjust the docs to drop the OT references.

src/index.mjs
```diff
@@ -150,14 +199,12 @@ export function listen(options) {
...
-      return Promise.reject(new Error('Cannot prefetch, Save-Data is enabled'));
-    }
-    if (/2g/.test(conn.effectiveType)) {
-      return Promise.reject(new Error('Cannot prefetch, network conditions are poor'));
-    }
+  checkConnection(conn = navigator.connection).catch(err => {
+    return Promise.reject(new Error('Cannot prefetch, '+err.message));
+  });
+  
+  if(toPrerender.size > 0 && !shouldPrerenderAndPrefetch){
```
addyosmani, line 206: Throughout the PR, I noticed there is inconsistent application of spacing when compared to the code being replaced. In this block for example this should be:

```
if (toPrerender.size > 0 && !shouldPrerenderAndPrefetch) {
```

Could you do a style pass just to make sure conventions used elsewhere in the repo are being followed?

src/index.mjs
```diff
@@ -175,3 +222,45 @@ export function prefetch(url, isPriority, conn) {
...
+    return Promise.reject(new Error('This Browser does not support Speculation Rules API. Falling back to Prefetch.'));
+  }
+  
+  // 2) whether spec rules is already defined (and with this we also covered when we have created spec rules before)
+  if (isSpecRulesExists()) {
+    return Promise.reject(new Error('Speculation Rules is already defined and cannot be altered.'));
+  }
+    
+  // 3) whether it's a same origin url,
+  try{
```
addyosmani, line 250: Similar style nit to the previous comment.

README.md
hadyan: Thanks Domenic and Addy, OT reference dropped in baeb7435c574fd21ef4ee28e42c65da13dee470a

src/prerender.mjs
hadyan: Thanks for pointing out Domenic, removed the extra argument and declared specScript as local variable.

src/index.mjs
hadyan: Thanks for catching this Addy, took another pass and fixed conventions inconsistencies.

src/index.mjs
hadyan: Also fixed with baeb7435c574fd21ef4ee28e42c65da13dee470a

src/index.mjs
hadyan: Thanks for catching this Domenic, should be fixed in 54eea9b07a7ef25c3b6102a36f9b8f6da6ae3c8c

src/index.mjs
```diff
@@ -36,6 +42,25 @@ function isIgnored(node, filter) {
     : (filter.test || filter).call(filter, node.href, node);
 }
 
+/**
+ * Checks network conditions
+ * @param  {NetworkInformation}  conn    The connection information to be checked
+ * @return {Promise}  reject is the constrainst are met or resolved otherwise
```
domenic, line 48: Return value documentation is no longer correct.

src/index.mjs
```diff
@@ -150,14 +199,13 @@ export function listen(options) {
...
 export function prefetch(url, isPriority, conn) {
-  if (conn = navigator.connection) {
-    // Don't prefetch if using 2G or if Save-Data is enabled.
-    if (conn.saveData) {
-      return Promise.reject(new Error('Cannot prefetch, Save-Data is enabled'));
-    }
-    if (/2g/.test(conn.effectiveType)) {
-      return Promise.reject(new Error('Cannot prefetch, network conditions are poor'));
-    }
+  let chkConn = checkConnection(conn = navigator.connection);
```
domenic, line 202: It's not clear why this is assigning to the `conn` parameter. Did you mean to set a default on line 201? Or just remove the `conn` parameter altogether? Or do `conn || navigator.connection`?

I guess this was already buggy...

src/index.mjs
```diff
@@ -175,3 +223,46 @@ export function prefetch(url, isPriority, conn) {
...
 }
+
+/**
+* Prerender a given URL
+* @param {String} url - the URL to fetch
+* @param {Object} [conn] - navigator.connection (internal)
+* @return {Object} a Promise
+*/
+export function prerender(urls, conn) {
+  let chkConn = checkConnection(conn = navigator.connection);
```
domenic, line 234: Same problem with assigning to a parameter here.

src/index.mjs
```diff
@@ -175,3 +223,46 @@ export function prefetch(url, isPriority, conn) {
...
+  if(chkConn instanceof Error) {
+    return Promise.reject(new Error('Cannot prerender, '+chkConn.message));
+  }
+  
+  // prerendering preconditions:
+  // 1) whether UA supports spec rules.. If not, fallback to prefetch
+  if (!hasSpecRulesSupport()) {
+    prefetch (urls);
+    return Promise.reject(new Error('This Browser does not support Speculation Rules API. Falling back to Prefetch.'));
```
domenic, line 243:
```suggestion
    return Promise.reject(new Error('This browser does not support the speculation rules API. Falling back to prefetch.'));
```

src/index.mjs
```diff
@@ -175,3 +223,46 @@ export function prefetch(url, isPriority, conn) {
...
+  // 3) whether it's a same origin url,
+  try{
+    [].concat(urls).filter(isSameOrigin).map(str => {
+      if (!toPrerender.has(str)) {
+        toPrerender.add(str);
+      }
+    });
+  } catch (e) {
+    return Promise.reject(new Error('Only same origin URL(s) are allowed: '+e));
+  }
```
domenic, line 260:
```suggestion
  // 3) whether it's a same origin url,
  for (const url of [].concat(urls)) {
    if (!isSameOrigin(url)) {
      return Promise.reject(new Error('Only same origin URLs are allowed: ' + url));
    }

    toPrerender.add(url);
  }
```

This change relies on my other change to make `isSameOrigin` return true or false.

src/prerender.mjs
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Checks if the given string is a same origin url
+ * @param {string} str - the URL to check
+ * @return {Boolean} true for same origin url
+ */
+export function isSameOrigin(str) {
+  if(typeof str === 'string' && window.location.origin !== new URL(str,window.location.origin).origin) {
+      throw str;
+  }
+  return true;
```
domenic, line 28: It would be clearer if this function just returned true or false, and left the exception stuff to the caller. So,

```suggestion
  return window.location.origin === (new URL(str, window.location.href)).origin;
````

File: src/prerender.mjs [hunk 13]
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/**
+ * Add a given set of urls to the speculation rules
+ * @param {Set} toPrerender - the URLs to add to speculation rules
+ * @return {Object} a promise
+ */
+export function addSpeculationRules(urlsToPrerender) {
```
domenic, line 36: It's not clear why this function returns a promise.

src/index.mjs
hadyan: Thanks, updated return value documentation.

src/index.mjs
hadyan: I have removed the `conn` parameter altogether as from my understanding it's not needed, but since it's been there in the earlier commits, perhaps @addyosmani can help to quickly double check to confirm it's ok, or is better to retain and use `conn || navigator.connection`? Thanks..

src/prerender.mjs
hadyan: Thanks for the feedback, updated the function's return value

addyosmani (approved):
LGTM

PR #272, opened 2022-11-29
Added support to observe a specific Anchor NodeList collection

gilbertococchi:
Took some inspiration from PR [183](https://github.com/GoogleChromeLabs/quicklink/pull/183) from @cajotafer.

Proposal: Allow to observe links specified via a specific NodeList collection of Anchor Elements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,7 @@ Quickstart:
 <script src="dist/quicklink.umd.js"></script>
 <!-- Initialize (you can do this whenever you want) -->
 <script>
-  quicklink.listen({
-    el: document.querySelectorAll('a.linksToPrefetch'),
-  });
+  quicklink.listen();
 </script>
 ```
 
@@ -57,9 +55,7 @@ For example, you can initialize after the `load` event fires:
 ```html
 <script>
 window.addEventListener('load', () =>{
-  quicklink.listen({
-    el: document.querySelectorAll('a.linksToPrefetch'),
-  });
+  quicklink.listen();
 });
 </script>
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -46,7 +46,9 @@ Quickstart:
 <script src="dist/quicklink.umd.js"></script>
 <!-- Initialize (you can do this whenever you want) -->
 <script>
-quicklink.listen();
+  quicklink.listen({
+    el: document.querySelectorAll('a.linksToPrefetch'),
```
addyosmani, line 50: (Don't hate me!) Let's restore the original / default behavior for this snippet :) I'm really happy for us to strongly encourage folks to set specific nodes to observe in a snippet right after this, but it would be a fundamental model shift to suggest you can't just drop in Quicklink and have it add value without some level of configuration.

gilbertococchi: Hi Addy, I would never hate you! 
I know, that's my bad, it's a radical change, I am only worried that unskilled developers would add Quicklink as is and they put down their Server without having a CDN, hitting too much Origin/Database etc.

What about keeping the original 0 friction use case and put this new method as the "suggested" method to restrict Prefetch to specific Links?

gilbertococchi:
@addyosmani, the quickstart example has been reverted, please review this when you have time.
Gil

dbniko:
хорош

gilbertococchi:
@rangerz, thanks for the feedback.

The options.ignores method would still continue to work when the library is initialized targeting an Element Ancestor of Links to prefetch.

Of course, when sending a direct NodeList the ignores parameter could not be taken in consideration but that's expected as the Developers intention should be explicit in that case.

Does it cover both needs in your opinion?

addyosmani (approved):

PR #282, opened 2023-02-16
feat: added minConnectionType to the options attribute

Luccatp:
a new options that gives developer the power of choosing which type of connection should be the minimum to prefetch or prerender the link, instead of always it being 3g.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,13 @@ Default: None
 
 An optional function to generate the URL to prefetch. It receives an [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) as the argument.
 
+#### options.minConnectionType
+
+Type: ` 'slow-2g' | '2g' | '3g' | '4g' | 'none'` <br>
+Default: '3g'
+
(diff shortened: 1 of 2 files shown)


XhmikosR:
Please rebase and make sure lint and tests pass.

XhmikosR:
I did my best to resolve the conflicts but it needs a proper look from the patch author.

The changes look breaking as they are and new tests are missing.

@Luccatp If you do not update the PR, we'll have to close it, but you can come back later and create a new one.

XhmikosR:
I fixed the conflicts and the breaking change but:

1. https://github.com/Luccatp/quicklink/blob/feat/minConnectionType/src/chunks.mjs should probably be updated too?
2. shouldn't `5g` be included too?
3. we need tests

Luccatp:
I will try to fix in this week time. 

The effectiveType uses only this values: https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType#value

Thank you for your collaboration and interest in my PR.

PR #308, opened 2023-03-30
babel: specify `loose: true` and `bugfixes: true`

XhmikosR:


addyosmani (approved):
LGTM, okay with this adjustment.

PR #312, opened 2023-03-30
Lint tweaks

XhmikosR:
The 3rd parameter in `prerender` and `prefetch` are moot after https://github.com/GoogleChromeLabs/quicklink/commit/383e4955b9264716a61ce11a7e1931584f0b59b4, but I'll let you verify @addyosmani.

If you think we should bring it back let me know.

This patch saves a few bytes.

BTW Happy to revert any changes if you prefer the old way. :)

Non-whitespace diff: https://github.com/GoogleChromeLabs/quicklink/pull/312/files?w=1

### DIFF
--- a/src/chunks.mjs
+++ b/src/chunks.mjs
@@ -56,23 +56,23 @@ function isIgnored(node, filter) {
  * @param {Function} [options.onError] - Error handler for failed `prefetch` requests
  * @param {Function} [options.prefetchChunks] - Function to prefetch chunks for route URLs (with route manifest for URL mapping)
  */
-export function listen(options) {
-  if (!options) options = {};
+export function listen(options = {}) {
   if (!window.IntersectionObserver) return;
 
-  const [toAdd, isDone] = throttle(options.throttle || 1/0);
-  const limit = options.limit || 1/0;
+  const [toAdd, isDone] = throttle(options.throttle || 1 / 0);
+  const limit = options.limit || 1 / 0;
 
   const allowed = options.origins || [location.hostname];
   const ignores = options.ignores || [];
 
   const timeoutFn = options.timeoutFn || requestIdleCallback;
 
-  const prefetchChunks = options.prefetchChunks;
+  const {prefetchChunks} = options;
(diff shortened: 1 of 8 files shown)


src/chunks.mjs
```diff
@@ -60,19 +60,20 @@ export function listen(options) {
...
+  const [toAdd, isDone] = throttle(options.throttle || 1 / 0);
+  const limit = options.limit || 1 / 0;
 
   const allowed = options.origins || [location.hostname];
   const ignores = options.ignores || [];
 
   const timeoutFn = options.timeoutFn || requestIdleCallback;
 
-  const prefetchChunks = options.prefetchChunks;
+  const {prefetchChunks} = options;
```
XhmikosR, line 71: Happy to revert any destructuring if you prefer the old way.

src/chunks.mjs
```diff
@@ -112,37 +113,38 @@ export function listen(options) {
...
+    if (/2g/.test(connection.effectiveType)) {
       return Promise.reject(new Error('Cannot prefetch, network conditions are poor'));
     }
   }
 
   // Dev must supply own catch()
   return Promise.all(
       [].concat(url).map(str => {
-        if (!toPrefetch.has(str)) {
+        if (toPrefetch.has(str)) return [];
```
XhmikosR, line 139: Unsure if this is better to be `undefined`?

addyosmani: I think that it should be fine to return `[]` as you currently have here

addyosmani (approved):

src/request-idle-callback.mjs
```diff
@@ -18,10 +18,10 @@
 const requestIdleCallback = window.requestIdleCallback ||
   function (cb) {
     const start = Date.now();
-    return setTimeout(function () {
+    return setTimeout(() => {
```
XhmikosR, line 21: @addyosmani I don't suppose this breaks the fallback for older supported browsers?

If so, we should revert these changes changes.

PR #318, opened 2023-03-31
Replace mocha with uvu

### LINKED ISSUE #317: Switch from mocha to uvu
uvu is written by @lukeed and it's pretty awesome and small.

I believe it should be a good replacement fit for this project's needs.

XhmikosR:
Fixes #317 

Non-whitespace diff: https://github.com/GoogleChromeLabs/quicklink/pull/318/files?w=1


addyosmani:
@XhmikosR if we're okay with resolving some of the conflicts (that are mostly a result of landing some of the other PRs), I'd be down with us merging this change in

XhmikosR:
Sure thing, I'll rebase tomorrow!

test/quicklink.spec.js
```diff
@@ -1,344 +1,358 @@
-const sleep = ms => new Promise(r => setTimeout(r, ms));
+const assert = require('assert').strict;
```
lukeed, line 1: Could also use `uvu/assert` in case you didn’t know :)

XhmikosR: I read about it, but haven't looked closer :)

I'll have a look again, thanks!

XhmikosR: BTW there's a failing test since the beginning when you added it. Any chance you could have a look?

https://github.com/GoogleChromeLabs/quicklink/blob/7d49abcec4de2b781e3435b3c62a3a9950fa5237/test/quicklink.spec.js#L232

PR #319, opened 2023-03-31
Move test fixtures to a subfolder

XhmikosR:
I personally find it cleaner this way, but it's an extra level. Let me know what you think :)

Non-whitespace diff: https://github.com/GoogleChromeLabs/quicklink/pull/319/files?w=1


addyosmani:
I'd be fine with moving them to a subdirectory tbh! If we can resolve the conflicts we can merge this in

PR #324, opened 2023-04-04
site: update devDependencies

XhmikosR:
Things look like they still work as before, but needs another pair of eyes :)


addyosmani (approved):
lgtm

PR #325, opened 2023-04-04
Replace prepare with `prepublishOnly`

XhmikosR:
@addyosmani I believe this is what was the original intention? Before, prepare ran after npm install which was annoying.

Feel free to close if the PR is wrong :)


addyosmani (approved):
LGTM

PR #327, opened 2023-04-04
site: Backport fixes from README

XhmikosR:
There might be another way to fix this, but I'm not too familiar with eleventy, so this fixes the `<br>`s shown for now :)

Another solution would be to switch these to a list in README and api.

Let me know which solution you prefer!


XhmikosR:
I pushed another patch which switches to `ul`. Let me know which one you prefer :)

addyosmani (approved):
LGTM

PR #341, opened 2023-04-07
approach.njk: use inline code in a couple of cases

XhmikosR:


site/src/approach.njk
```diff
@@ -36,7 +36,7 @@ What this means for Quicklink is that the library will continue to work for same
 
 ## Session Stitching
 
-Cross-origin prefetching (e.g a.com/foo.html prefetches b.com/bar.html) has a number of limitations. One such limitation is with session-stitching. b.com may expect a.com's navigation requests to include session information (e.g a temporary ID - e.g b.com/bar.html?hash=<>&timestamp=<>), where this information is used to customize the experience or log information to analytics.
+Cross-origin prefetching (e.g `a.com/foo.html` prefetches `b.com/bar.html`) has a number of limitations. One such limitation is with session-stitching. `b.com` may expect `a.com`'s navigation requests to include session information (e.g a temporary ID - e.g `b.com/bar.html?hash=<>&timestamp=<>`), where this information is used to customize the experience or log information to analytics.
```
XhmikosR, line 39: Note that `&` is still not properly escaped in the generated HTML, but I guess we can live with that for now :)

PR #352, opened 2023-04-19
Update Prerender Support

### LINKED ISSUE #348: Fix Prerender Support
Now QuickLink limits Prerender to 1 URL and to Same Origin only.

- [ ]  1 URL to Prerender limit should be removed
- [ ] Same Origin should be removed or it should include Same-Site Cross-Origin use case as well

@hadyan  @addyosmani FYI

hadyan:
Update Prerender limit and checks based on the updated UA Prerender limit and feature support:

1. Remove 1 Prerender limit (with [support for multiple Prerenders](https://docs.google.com/document/d/1Cp4KK6lVftKcsrrlg5F5yOfXxZZW_jilcOAfbRTIA64/))
2. Drop pre-existing speculation rules check since multiple speculation rules declaration now appends to the existing speculation rules
3. Drop same origin check (with [support for same-site cross origin](https://docs.google.com/document/d/1WsDYA8NMCSwsK8dXCKdajdAd3ZcQUu9w1eoe0hEB_nU/))

Fixes #348


XhmikosR:
If this is not a breaking change it could go in the next minor version, otherwise let's land it after that. Alternatively, we could make the next version a major bump. I'll let you guys decide :)

site/src/api.njk
```diff
@@ -173,7 +180,7 @@ Returns: `Promise`
 
 One or many URLs to be prerendered.
 
-> **Note:** As prerendering using Speculative Rules API only supports same-origin at this point, only same-origin urls are accepted. Any non same-origin urls will return a rejected Promise.
+> **Note:** Speculative Rules API supports same-site cross origin Prerendering with [opt-in header](https://docs.google.com/document/d/1WsDYA8NMCSwsK8dXCKdajdAd3ZcQUu9w1eoe0hEB_nU/).
```
addyosmani, line 183: Minor: it would be nice if we could put this behind a URL shortener vs. including a completely Google Docs URL in source. I created https://bit.ly/ss-cross-origin-pre for this purpose if you'd like to use that.

addyosmani (approved):
I've read through the full implementation and changes LGTM.

addyosmani:
> If this is not a breaking change it could go in the next minor version, otherwise let's land it after that. Alternatively, we could make the next version a major bump. I'll let you guys decide :)

I'd be comfortable with this change going into the next minor fwiw :)

site/src/api.njk
hadyan: Thanks for the review, suggestion and helping to create the short URL. It has been updated in the latest minor commit.

PR #356, opened 2023-04-19
site: use page title in `title`

XhmikosR:


site/src/_includes/layouts/head.njk
```diff
@@ -1,6 +1,6 @@
 <head>
   <meta charset="utf-8">
-  <title>{{ title if title else site.title }}</title>
+  <title>{{ [title, "|", site.title] | join(" ") }}</title>
```
XhmikosR, line 3: Not super familiar with nunjunks, let me know if there's a better/simpler way

addyosmani (approved):

PR #358, opened 2023-04-19
site: switch to Sass, add autoprefixer

XhmikosR:
Notes:

* I didn't convert the CSS syntax to Sass, just the comments for now
* I included all Sass files in main, it probably isn't optimal, but it is simpler. I suppose we could try to make a critical.scss and preload that only. Let me know which files to include if you want to go with the latter approach. I also tried to use critical itself, but I'm having trouble making it to work in this project.
* Unsure what's the situation with the 2 github-markdown files; I included them both in main

/CC @addyosmani


XhmikosR:
I'm going to land this as is and we can revisit later. Basically we should have one final CSS final and critical set up so that we generate the critical CSS automatically, but I couldn't make it work.

PR #359, opened 2023-04-19
site: add Stylelint support

XhmikosR:


site/.stylelintignore
```diff
@@ -0,0 +1,5 @@
+**/*.min.css
+**/dist/
+**/vendor/
+/build/
+prism.css
```
XhmikosR, line 5: Later, I plan to move the vendor styles to a subfolder

PR #361, opened 2023-04-19
site: add a sitemap and social tag fixes

XhmikosR:
This is a sort of naive way but it seems to do the job.

~~EDIT: draft because I want to reuse the url variable in all places while at it.~~

### DIFF
--- a/site/src/_data/site.json
+++ b/site/src/_data/site.json
@@ -3,6 +3,7 @@
   "subtitle": "Instant next-page navigations",
   "description": "Faster subsequent page-loads by prefetching in-viewport links during idle time.",
   "logo": "/assets/images/logos/quicklink.svg",
+  "url": "https://getquick.link",
   "quicklinkGithubURL": "https://github.com/GoogleChromeLabs/quicklink",
   "quicklinkVersion": "2.3.0",
   "quicklinkSizeLimit": "1KB",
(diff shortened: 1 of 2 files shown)


site/src/_includes/layouts/head.njk
```diff
@@ -1,10 +1,10 @@
 <head>
   <meta charset="utf-8">
   <title>{{ [title, "|", site.title] | join(" ") }}</title>
-  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="description" content="{{ description }}">
+  {% set page_url = [site.url, page.url | url] | join -%}
   <meta name="twitter:card" content="summary">
-  <meta name="twitter:site" content="https://getquick.link">
```
XhmikosR, line 7: @addyosmani I removed this because it's supposed to be a Twitter handler. Is there a Twitter account for Quicklink? If not, I guess the removal is OK.

site/src/assets/images/og-image.png
XhmikosR, line 1: Note that I used the transparent image, but we might as well go with the white bg one to make sure it looks good in dark mode.

XhmikosR:
I will move the other patches to another PR.

PR #368, opened 2023-04-21
request-idle-callback.mjs: switch to a named export

XhmikosR:
**WARNING**: Can this be considered a breaking change? AFAICT this is not public API, but if it's considered a BC let's postpone it for v3.0.0.


addyosmani:
I would agree that considering this a breaking change for the next major may make sense. Created a label for us to distinguish which changes should be saved for 3.0 vs can land earlier.

addyosmani (approved):
LGTM but as noted let's wait for 3.0

PR #371, opened 2023-04-21
site: switch to @11ty/eleventy-navigation

XhmikosR:
I feel like this approach is the proper one without us having to maintain anything outside of the front matter.

BTW @addyosmani how about setting up Netlify previews? If you make an account, you can add me as an admin on Netlify temporarily so that I set it up, and then you can remove me.

### DIFF
--- a/site/.eleventy.js
+++ b/site/.eleventy.js
@@ -2,9 +2,10 @@
 
 'use strict';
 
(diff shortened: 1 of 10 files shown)


XhmikosR:
BTW @addyosmani how about setting up Netlify previews? If you make an account, you can add me as an admin on Netlify temporarily so that I set it up, and then you can remove me.

PR #373, opened 2023-04-21
site: switch to using `EleventyHtmlBasePlugin`

XhmikosR:
This should be more safe in case the base is changed and it's less stuff to handle manually in templates.

See https://www.11ty.dev/docs/plugins/html-base/

Might need to rebase #361 after this one, or this one after #361

### DIFF
--- a/site/.eleventy.js
+++ b/site/.eleventy.js
@@ -2,10 +2,11 @@
 
 'use strict';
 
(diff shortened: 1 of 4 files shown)


XhmikosR:
Marking it as draft until we are done with the other non-draft PRs first.

PR #374, opened 2023-04-21
site: fix `og:url` and remove `twitter:site`

XhmikosR:
Split from #361

### DIFF
--- a/site/src/_includes/layouts/head.njk
+++ b/site/src/_includes/layouts/head.njk
@@ -1,17 +1,16 @@
 <head>
   <meta charset="utf-8">
   <title>{{ title + " | " + site.title if title else site.title }}</title>
-  <meta name="viewport" content="width=device-width,initial-scale=1">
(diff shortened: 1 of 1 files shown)


site/src/_includes/layouts/head.njk
```diff
@@ -1,10 +1,9 @@
 <head>
   <meta charset="utf-8">
   <title>{{ title + " | " + site.title if title else site.title }}</title>
-  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="description" content="{{ description }}">
   <meta name="twitter:card" content="summary">
-  <meta name="twitter:site" content="https://getquick.link">
```
XhmikosR, line 7: @addyosmani I removed this because it's supposed to be a Twitter handler. Is there a Twitter account for Quicklink? If not, I guess the removal is OK.

PR #381, opened 2023-04-22
site: switch to prism tomorrow theme

XhmikosR:
It's better for a11y, although it probably doesn't look too nice. :/

I tested a few more official themes and this was one of the few themes that didn't have color contrast issues.

image


addyosmani:
I'm happy for us to optimize for better color contrast and we can always switch to a prettier version if we find that there's a theme that matches all of our criteria.

addyosmani (approved):
LGTM

PR #388, opened 2023-06-16
docs: set site URL dynamically for Netlify

XhmikosR:
Also, disable robots crawling on Netlify


XhmikosR:
@addyosmani LGTY? I went with the approach with one data file.

Issue #387, opened 2023-05-09
No prefetching occurs for duplicate links


Pimm:
**Describe the bug**

When a link appears on the page twice ‒ once inside and once outside the viewport ‒ _and_ a delay has been set, the link will not cause prefetching/prerendering.

**To Reproduce**

```html
<!doctype html>
<html>

<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <link rel="stylesheet" href="main.css">
</head>

<body>
  <div class="screen">
    <a href="1.html">Link</a>
  </div>
  <a href="1.html">Same link</a>
  <script src="../../dist/quicklink.umd.js"></script>
  <script>
    quicklink.listen({ delay: 1000 });
  </script>
</body>

</html>
```

**Expected behavior**

The page to be prefetched, as at least one link to it is in-viewport.

---

The `IntersectionObserver` detects that the first link is in-viewport, and [appropriately pushes](https://github.com/GoogleChromeLabs/quicklink/blob/9f2949034e3e30269447d8758182632ced484d5e/src/index.mjs#L123) into the `hrefsInViewport` array. It immediately afterwards detects that the second link is not in-viewport, and thus [removes the freshly added item](https://github.com/GoogleChromeLabs/quicklink/blob/9f2949034e3e30269447d8758182632ced484d5e/src/index.mjs#L162) from `hrefsInViewport`. `hrefsInViewport` ‒ as the name suggests ‒ contains hrefs, thus links with equal hrefs are indistinguishable in this context.

I am willing to submit a PR, but would like to make I'm not the only one who thinks this is unexpected.

One possible solution would be storing some identifier for a link that is unique to it, rather than the href (which is not necessarily unique). Such an identifier could be generated like so:
```javascript
var identify = function(ids, node) {
  var id = ids.get(node);
  if (!id) {
    ids.set(node, id = {});
  }
  return id;
}.bind(undefined, new WeakMap());
```

PR #403, opened 2023-09-25
Fix duplicate links not being preloaded

daun:
- Index visible links by href first, by element second
- Fixes issue #387 where duplicate links with identical hrefs won't preload at all if one link is visible and another one isn't


XhmikosR:
@hadyan could you review this PR please? :)

XhmikosR:
@addyosmani please review this PR too

Issue #402, opened 2023-09-17
Avoid Triggering Requests while offline


gilbertococchi:
QuickLink should avoid triggering Network requests when the device is offline.

It should be relatively easy to fix this by adding this code in the `checkConnection` function.

```
if("onLine" in navigator && !navigator.onLine) {
  return new Error('Device is offline');
}
```

PR #411, opened 2024-01-09
Avoid Triggering Requests while offline #402

HIMU-2001:


XhmikosR:
@hadyan could you review this PR please? :)

PR #429, opened 2024-06-26
Update polyfill link

iosh:
The usage of polyfill.io is not secure. Please refer to the following link for details : https://sansec.io/research/polyfill-supply-chain-attack

So switch to Fastly. More information can be found here : https://community.fastly.com/t/new-options-for-polyfill-io-users/2540

PR #432, opened 2024-07-31
[readme] Remove reference to polyfill.io

Daniel15:
polyfill.io is blocked by most security software and PiHole/AdGuardHome blocklists now. Intersection observer support is at 97% according to CanIUse, so it's not needed in most cases.


addyosmani (approved):
lgtm

XhmikosR:
We should make the same change in /site, see #429

PR #438, opened 2025-03-22
Update safe dependencies

XhmikosR:


XhmikosR:
@lukeed sorry for CC'ing you, do you have any idea why these CI errors started to appear? They happen on main branch too.

```
 quicklink.spec.js
TypeError: Cannot read properties of undefined (reading 'close')
    at /home/runner/work/quicklink/quicklink/test/quicklink.spec.js:28:22
    at Number.runner (/home/runner/work/quicklink/quicklink/node_modules/uvu/dist/index.js:91:29)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Module.exec (file:///home/runner/work/quicklink/quicklink/node_modules/uvu/dist/index.mjs:141:33)
    at async Module.run (file:///home/runner/work/quicklink/quicklink/node_modules/uvu/run/index.mjs:12:2)
    at async /home/runner/work/quicklink/quicklink/node_modules/uvu/bin.js:26:5
```

I don't get these errors on my local dev Windows 10 VM.

addyosmani:
Thank you very much for digging in to debug the CI errors. I was seeing these locally as well and couldn't quite get to the bottom of what was happening there. Looking at the specific changes we've been landing those did not appear to be the root cause and I think this may have been related to particular dependencies but I could not be sure just yet.

XhmikosR:
It doesn't look like it's the deps. Something is different on Linux, but I don't get what's wrong. On my Windows VM the errors don't appear, I will try on Linux tomorrow.

I still want to use uvu because it's lightweight, but worse case scenario we switch test library or switch to Node.js native test runner which works well but needs Node.js 18 and the colors are broken on Windows native cmd (the latter is what bugs me most, but I guess I'll live with it :))

addyosmani:
> worse case scenario we switch test library or switch to Node.js native test runner which works well but needs Node.js 18 and the colors are broken on Windows native cmd (the latter is what bugs me most, but I guess I'll live with it :))

Thanks for confirming its likely not the deps. I also appreciate the fallback options you outlined :)

Issue #442, opened 2025-05-12
Refactor addSpeculationRules to Return Individual DOM References for Removability


giorgiopellegrino:
**Description**

The addSpeculationRules function should support dynamic addition and removal of speculation rules in alignment with recent platform capabilities.

**Motivation**

The current implementation adds all URLs in a single <script type="speculationrules"> element. However, with support for removing speculation rules based on viewport visibility, it's more effective to:

- Insert one speculation rule per URL.
- Return a structured response linking each URL to its corresponding DOM element, allowing the caller to remove individual rules without querying the DOM.

**Changes to introduce**

- For each URL, a separate speculation rule script should be created and appended to the document.
- The function should return an array of Maps in the format:

```
{
  "/page1": HTMLScriptElement,
  "/page2": HTMLScriptElement,
  ...
}
```

**Example Usage**

```
const urls = addSpeculationRules(["/page1", "/page2"], "eager");
document.head.removeChild(urls["/page1"]);
```
This approach improves clarity and simplifies DOM cleanup when rules are no longer needed.

cc @gilbertococchi @addyosmani

gilbertococchi:
Thanks Giorgio, @tunetheweb @hadyan, @domenic FYI

That sounds like a great plan, I was thinking it that it might be good reference these HTMLScriptElements (the SpeculationRules tags) to the IntersectionObserverTarget itself to make the link between the Link entering or quitting the viewport and the element needed to be added or removed.

Do you agree?

giorgiopellegrino:
Exactly, @gilbertococchi . My goal is to modify the addSpeculationRules(urlsToPrerender, eagerness) function, located in prerender.mjs, so that when it is invoked by the listen(el) method in index.mjs, the corresponding speculation rules can be removed from the DOM when the associated element exits the IntersectionObserver.
@tunetheweb If you believe it would be more appropriate to manage the map using WeakRef entries, I would be in agreement with that approach. Thank you.
Giorgio

PR #446, opened 2025-05-20
Improve speculation rules handling based on element visibility

giorgiopellegrino:
**Pull Request Description:**
This pull request addresses the issue raised in https://github.com/GoogleChromeLabs/quicklink/issues/442(https://github.com/GoogleChromeLabs/quicklink/issues/442) by enhancing how *speculation rules* are managed based on the visibility of DOM elements.

Specifically, speculation rules are now **added only when the target element is within the viewport** and **removed when it exits the viewport**. This change enables developers to safely use `quicklink.listen(el, { prerender: true, eagerness: 'immediate' })` on mobile devices without the risk of hitting the browser-imposed limit of 10 prerendered pages.

This implementation improves resource management and ensures a more efficient and predictable prerendering behavior.

Thanks,

Giorgio Pellegrino.

cc @gilbertococchi


src/index.mjs
```diff
@@ -131,19 +132,24 @@ export function listen(options = {}) {
           // Do not prefetch if not found in viewport
           if (!hrefsInViewport.includes(entry.href)) return;
 
-          observer.unobserve(entry);
```
gilbertococchi, line 134: Can you please elaborate why you removed unobserve from here?

gilbertococchi (commented):
This is great Giorgio!

I've run some tests and it looks like a great start!

Would you be open to have a testing HTML page added to verify the behaviour and also add some documentation changes that explains what this combination will work on QuickLink?

XhmikosR:
CI fails, please make sure tests are valid.

gilbertococchi (approved):
Thanks Giorgio, all the changes here looks good now.

gilbertococchi:
Thanks @giorgiopellegrino for adding the documentation changes, can you please add them as part of this PR?

Context: https://github.com/giorgiopellegrino/quicklink/commit/f522eef5f70a11d74c867d866899eac7ad718e1c

src/index.mjs
```diff
@@ -102,6 +102,7 @@ export function listen(options = {}) {
   const ignores = options.ignores || [];
   const delay = options.delay || 0;
   const hrefsInViewport = [];
+  specRulesInViewport = new Map();
```
westonruter, line 105: Why is this missing `const`?

```suggestion
  const specRulesInViewport = new Map();
```

src/prerender.mjs
```diff
@@ -19,23 +19,52 @@
...
+  }
+
+  return specMap;
+}
+
+/**
+ * Removes a speculation rule script associated with a given URL
+ * @param {Map<string, HTMLScriptElement>} specMap - Map of URLs to their script elements
+ * @param {string} url - The URL whose speculation rule should be removed
+ * @return {Map<string, HTMLScriptElement>|Object} The updated map after removal or Error Object
```
westonruter, line 55:
```suggestion
 * @return {Map<string, HTMLScriptElement>|Error} The updated map after removal or Error Object
```

gilbertococchi (approved):

src/index.mjs
```diff
@@ -131,19 +132,26 @@ export function listen(options = {}) {
           // Do not prefetch if not found in viewport
           if (!hrefsInViewport.includes(entry.href)) return;
 
-          observer.unobserve(entry);
+          if (!shouldOnlyPrerender && !shouldPrerenderAndPrefetch) observer.unobserve(entry);
```
westonruter, line 135: As a best practice I think the statement should be wrapped in braces:

```suggestion
          if (!shouldOnlyPrerender && !shouldPrerenderAndPrefetch) {
            observer.unobserve(entry);
          }
```

src/index.mjs
```diff
@@ -245,6 +256,8 @@ export function prefetch(urls, isPriority, checkAccessControlAllowOrigin, checkA
 * @return {Object} a Promise
 */
 export function prerender(urls, eagerness = 'immediate') {
+  urls = [].concat(urls);
+
```
westonruter, line 260: I wonder why this is necessary? Is it to create a copy to so that if the underlying object is modified during some async iteration to prevent unexpected behavior? I don't see this as being the case, since this is not an async function and below it just loops over the `urls` to add each to the `toPrerender` Set. 

I think this can just be removed.

```suggestion
```

README.md
```diff
@@ -126,6 +126,14 @@ A "reset" function is returned, which will empty the active `IntersectionObserve
...
 
 > **Note:** The prerendering mode (when this option is set to true) will fallback to the prefetching mode if the browser does not support prerender.
+> Once the element exits the viewport, the speculationrules element is removed from the DOM. This approach makes it possible to exceed the limit of 10 prerenders imposed for the eagerness: 'immediate' and eagerness: 'eager' settings.
+
+#### options.eagerness
+
+- Type: `String`
+- Default: `immediate`
+
+Determines the mode to be used for prerendering specified within the speculationrules.
```
westonruter, line 136:
```suggestion
Determines the mode to be used for prerendering specified within the speculation rules.
```

README.md
```diff
@@ -281,6 +289,13 @@ One or many URLs to be prerendered.
 
 > **Note:** Speculative Rules API supports same-site cross origin Prerendering with [opt-in header](https://bit.ly/ss-cross-origin-pre).
 
+#### eagerness
+
+- Type: `String`
+- Default: `immediate`
+
+Determines the mode to be used for prerendering specified within the speculationrules.
```
westonruter, line 297:
```suggestion
Determines the mode to be used for prerendering specified within the speculation rules.
```

README.md
```diff
@@ -126,6 +126,14 @@ A "reset" function is returned, which will empty the active `IntersectionObserve
 Whether to switch from the default prefetching mode to the prerendering mode for the links inside the viewport.
 
 > **Note:** The prerendering mode (when this option is set to true) will fallback to the prefetching mode if the browser does not support prerender.
+> Once the element exits the viewport, the speculationrules element is removed from the DOM. This approach makes it possible to exceed the limit of 10 prerenders imposed for the eagerness: 'immediate' and eagerness: 'eager' settings.
+
+#### options.eagerness
+
+- Type: `String`
+- Default: `immediate`
```
westonruter, line 134: I wonder if `immediate` is too eager? Why not `eager`? It's true that today they behave the same, but the intention is that `eager` can could potentially be more nuanced in the future, [per](https://developer.chrome.com/blog/speculation-rules-improvements#:~:text=eager%3A%20This%20currently%20behaves%20identically%20to%20the%20immediate%20setting%2C%20but%20in%20future%2C%20we%20are%20looking%20to%20place%20this%20somewhere%20between%20immediate%20and%20moderate.) @tunetheweb:

> `eager`: This currently behaves identically to the `immediate` setting, but in future, we are looking to place this somewhere between `immediate` and `moderate`.

Since there could be a lot of links in the viewport without any signal about whether they are going to be navigated to (other than them being in the viewport), it would seem like `immediate` is too aggressive.

tunetheweb: `eager` is likely to be replaced with a short (or even 0) hover time—basically moderate with minimal delay. So it would still require a user interaction. I'd imagine if/when that happens we might reduce the limit to be 2 similar to `moderate` and `conservative`.

The intent for this logic is particularly intended for mobile, where there is no hover, so `immediate` seems like the right solution.

src/index.mjs
gilbertococchi: Hi @westonruter, this line was already present in the existing Speculation Rules implementation, so not really related to this PR.

It was done this way to ensure that URLs would be treated as array of strings.

@giorgiopellegrino to follow up in case an action is needed

README.md
westonruter: Let's keep it `immediate`

src/index.mjs
westonruter: I see. The arg is `String | String[]`, so this does essentially cast to an array.

westonruter (approved):
LGTM!

addyosmani:
A big thank you to everyone for the very thorough reviews on these changes and to @giorgiopellegrino for the additions. These LGTM to be merged and will be included in the next release.

addyosmani (approved):

XhmikosR:
This shouldn't have landed as is; there's a syntax error as shown by CodeQL :/

giorgiopellegrino:
Hi @XhmikosR,
thank you very much for spotting this change and for using the set method on maps instead of assignment to ensure the instance always remains unique.
Thanks again,
Giorgio Pellegrino.

giorgiopellegrino:
> @giorgiopellegrino Please update the site docs for the new options you added here :)

I've updated the GitHub documentation. Are you planning to update the API page on the website as well (https://getquick.link/api/)? How does that work?

PR #444, opened 2025-05-13
Modernize code and improve readability

XhmikosR:

PR #450, opened 2025-06-06
Minor site tweaks

XhmikosR:


XhmikosR:
@addyosmani could you review this PR, #444, then #403 and last #449?

I think we could then release a new minor/patch version.

addyosmani:
@XhmikosR Yes! Happy to. Could I ask for a check that those 3 are rebased against master just so the conflict resolutions reflect your opinions as well?

PR #461, opened 2025-12-06
Update GitHub actions

XhmikosR:


XhmikosR:
@addyosmani any news about allowing dependabot on this repo? If it's a matter of making our dependabot config stricter, we can adapt it with cooldown period. But not updating the actions it's cumbersome and doing it by hand doesn't scale...

PR #464, opened 2025-12-06
tests: mock external requests

XhmikosR:
@addyosmani WDYT? This should help with the tests flakiness since we no longer depend on external URLs.

Note that I used AI to help out.


XhmikosR:
I'm going to merge this @addyosmani but I'd appreciate if you could look at the other PRs.

PR #465, opened 2025-12-06
site: use quicklink from dist and switch to npm workspaces

### LINKED ISSUE #389: site: use the files from dist folder instead of the CDN
Not sure how to load the dist folder which is one folder up from site.

This should make sure PR previews use the latest code plus we should be able to add SRI hashes later (see #336).

XhmikosR:
Closes #389

TODO:

- [ ] Adapt Netlify after merging this so that we don't run double builds
- [x] Watch doesn't trigger if dist changes

Preview: <https://deploy-preview-465--google-chrome-labs-quicklink.netlify.app/>


addyosmani:
Now that #463 landed, this'll want a rebase on top of the new rollup build. Looks like the only open item on your list is the Netlify double-build — happy to sort that on the Netlify side once this is ready. Flip it out of draft whenever and I'll review promptly. 👍

XhmikosR:
I rebased it. It needs further testing. I couldn't find another way to make things work, so do let me know how it works for you or if you have any other ideas.

XhmikosR:
Actually, let's put this on hold and look at my other PRs first. This needs some work. I'll drop clipboard since nowadays browsers have native support.

addyosmani:
Heads-up that the hosting moved under this PR: getquick.link is now served from Netlify (the old Firebase project became inaccessible, so I migrated it). That makes your "adapt Netlify so we don't run double builds" TODO the real blocker here rather than a nice-to-have — once the site consumes quicklink from `dist` via the workspace, we'll want the Netlify build to run the workspace build once and publish `site/build`. The dist-over-CDN direction is right regardless; it drops the runtime dependency on jsDelivr and on whatever version happens to be published. Ping me when you flip it out of draft and I'll wire up the Netlify build command / a deploy action.

XhmikosR:
Can you add me as an administrator on Netlify so that i set thibgs up?
Previously, we were using my account for previews.

PR #480, opened 2026-05-02
Replace dead Glitch demos with self-hosted equivalents

### LINKED ISSUE #469: Demo links are broken
All the links in demo page are broken: https://getquick.link/demo. Please check

Image

addyosmani:
Closes #469.

Glitch shut down at the end of 2025, which broke every demo link on
https://getquick.link/demo and https://getquick.link/measure, and most of
the links in the repo's README. The original demo source lived on Glitch
and is gone, so this PR rebuilds the demos from scratch and serves them
from the same Eleventy site, under `/demos/`.

### What's in the new demos

- **`/demos/mini-ecomm/`** - vanilla 8-product store with a listing page
  and product detail pages. No Quicklink, so each click hits the
  network.
- **`/demos/mini-ecomm-quicklink/`** - same store, with `quicklink.listen()`
  loaded from the jsdelivr UMD bundle on the listing page. Drop-in
  comparison to the unoptimized variant for the demo and measure pages.
- **`/demos/spa/`** - a tiny vanilla-JS single-page app with a hand-rolled
  hash router and one ES-module chunk per route. Quicklink is wired up
  with an `hrefFn` that maps each nav link to its chunk URL, so it
  prefetches `routes/blog.js` etc. instead of an HTML document. This
  replaces the create-react-app demo and avoids needing a build step.

All three demos are static, so they ride the existing Eleventy →
Firebase deploy with no extra config - just an `addPassthroughCopy` and
an `.eleventyignore` entry.

### Other changes

- Updated `site/src/demo.njk` and `site/src/measure.njk` to link to the
  new `/demos/...` paths. The WebPageTest script examples now use
  absolute `getquick.link` URLs and click `a.card` (matching the new
  markup) instead of the first `<a>` on the page.
- Updated the README's Demo section: dropped the four "Glitch demos"
  bullets and replaced them with three live links plus a pointer to
  `site/src/demos/` for source.
- The legacy `demos/news/`, `demos/news-workbox/` and `demos/spa/`
  README-only folders now point at the self-hosted replacements with a
  short note about the Glitch history, so any external links to those
  folders still land somewhere useful.

### How to verify

1. `cd site && npm run build && npm run lint` - both pass.
2. `npx serve build` (or any static server) and visit
   `/demos/mini-ecomm/`, `/demos/mini-ecomm-quicklink/`, and `/demos/spa/`.
3. On the optimized mini-ecomm and the SPA, open DevTools → Network
   and confirm that `quicklink` shows up as the **Initiator** with
   **Lowest** priority for in-viewport links.
4. Throttle to Fast 3G and click around - the optimized variants should
   serve from prefetch cache.


XhmikosR:
Thanks for the PR! BTW, if I'm not CC'ed I don't get notified.

I'll have a closer look later, but it looks good so far. I'd only change the emdashes (`—`) with normal hyphens. AI tends to always use Unicode characters.

XhmikosR:
Let's land it and we can tweak it later if needed.

XhmikosR (approved):

Issue #204, opened 2020-11-05
The README file in this repo has some bad links - [404:NotFound]


MrCull:
**Describe the bug**
The markup version of the readme that is displayed in the main page for this repo contains the following links:
Status code [404:NotFound] - Link: https://keyword-2-ecd7b.firebaseapp.com/
Status code [404:NotFound] - Link: https://github.com/GoogleChromeLabs/quicklink/blob/master/#custom-ignore-patterns

(The links in the readme’s markdown may appear different to the link above extracted from the markup)

**To Reproduce**
Find the links under "demo" & "custom ignore patterns list"

**Expected behavior**
Links should work!

**Version:**
n/a

**Additional context, screenshots, screencasts**

The bad link was found by a tool I very recently created as part of an new experimental hobby project: https://github.com/MrCull/GitHub-Repo-ReadMe-Dead-Link-Finder
I (a human) verified that this link is broken (for me) and have manually logged this Issue (i.e. this Issue has not been created by a bot).
If this has been in any way helpful then please consider giving the above Repo a Star.

If you have any feedback on the information provided here, or on the tool itself, then please feel free to share your thoughts and pass on the feedback.

passionInfinite:
@addyosmani Can I please take this issue for a start?

XhmikosR:
@addyosmani AFAICT only the demo link is broken. Should we just link to `https://getquick.link/demo/`?

sankalpsthakur:
I'd like to take this — will fix the README 404 links.

sankalpsthakur:
Withdrawing for now — coordinating another quicklink docs PR first; will revisit if still open.

SatvikMishra08:
I'll take this — fixing the broken Research demo link in the README (keyword-2-ecd7b.firebaseapp.com → https://getquick.link/demo/).

PR #487, opened 2026-06-13
fix(readme): repair broken links in README

abhay8463:
Fixes issue #204 where the README file had broken links.

## Changes
- Replace broken Firebase demo link (https://keyword-2-ecd7b.firebaseapp.com/) with correct getquick.link domain
- Fix GitHub self-reference in "Ad-related considerations" section to use relative path (#custom-ignore-patterns) instead of absolute URL

## Details
Two broken links were identified and fixed:
1. In the Research section: Changed the demo link from Firebase hosting to the correct getquick.link domain
2. In Additional notes > Ad-related considerations: Changed the self-referencing link from absolute GitHub URL to relative path within the document

These changes ensure all links in the README are functional and point to the correct resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ By default, these errors are silently ignored.
 #### options.hrefFn
 
 - Type: `Function`
-- Default: None
+- Default: `None`
 
 An optional function to generate the URL to prefetch. It receives an [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) as the argument.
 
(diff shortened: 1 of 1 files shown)


XhmikosR (changes requested):
You have made unrelated changes to the link change.

hiSandog:
The README link fixes look straightforward. One thing worth double-checking is whether the replacement `getquick.link` demo is meant to be a stable project-owned URL rather than another temporary demo host. For the self-reference, the relative anchor is a good improvement because it will keep working across forks and branches.

PR #490, opened 2026-07-22
Skip prefetching anchors with the download attribute

### LINKED ISSUE #181: Do not prefetch links with `download` attribute
**Is your feature request related to a problem? Please describe.**
Currently all links are prefetched including once intended for downloadable assets. Since these links could point to large payloads (imagine a multi megabyte PDFs or zip archives) that's quite a waste of bandwidth.

I'd argue excluding links which have the `download` attribute would be a sensible default.

**Describe the solution you'd like**
```js
document.querySelectorAll('a:not([download])')
```

Thank you :)

sankalpsthakur:
## Summary
- Do not observe / prefetch `<a download>` links by default (boolean or named `download` attribute)
- Applies in both `listen` (`src/index.mjs`) and the chunks path (`src/chunks.mjs`)
- Adds a fixture + puppeteer test and a short README note under Custom Ignore Patterns

Fixes #181

## Why this matters
Download links often target large binaries (PDFs, archives, exports). Prefetching them wastes bandwidth and cache space on paths the user may never follow — especially on metered or Save-Data connections where quicklink already throttles prefetch. Treating `download` like other built-in ignore patterns (external origins, `rel=nofollow`, etc.) keeps prefetch focused on navigational routes.

## Test plan
- [x] `npm test` (19/19 passing)
- [x] `npm run size` (within limits)
- [x] `npm run lint`
- [x] Confirm CLA check passes for @sankalpsthakur

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,8 @@ quicklink.listen({
 
 ### Custom Ignore Patterns
 
+By default, quicklink skips anchors that have a [`download`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download) attribute, since those often point at large binary payloads.
+
 These filters run _after_ the `origins` matching has run. Ignores can be useful for avoiding large file downloads or for responding to DOM attributes dynamically.
 
 ```js
--- a/src/chunks.mjs
+++ b/src/chunks.mjs
@@ -106,7 +106,7 @@ export function listen(options = {}) {
   });
 
   timeoutFn(() => {
-    // Find all links & Connect them to IO if allowed.
+    // Find all links & connect them to IO if allowed.
     // Skip anchors with the `download` attribute (large payloads).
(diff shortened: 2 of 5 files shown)


sankalpsthakur:
CLA signed for @sankalpsthakur / sankalphimself@gmail.com. `cla/google` is green.

addyosmani (approved):
Thank you — this is the right default. A `download` anchor is an explicit signal that the link isn't a navigation, so prefetching it never pays off. What sold me is that you handled both paths: the `:not([download])` selector *and* the explicit `hasAttribute('download')` guard for the `options.el` NodeList case — that second one is easy to miss and would've leaked download links right through. Covering both the boolean and the named `download="report.pdf"` forms in the test is exactly what I'd want here. Merging.

PR #495, opened 2026-08-04
site: Lighthouse fixes

XhmikosR:


addyosmani (approved):
Thank you, these are all sensible wins. The `rel="dns-preconnect"` → `rel="preconnect"` fix is a genuine one — `dns-preconnect` isn't a valid link type, so that hint was silently doing nothing. Deferring the clipboard script and underlining in-body markdown links (not leaning on color alone) are the right calls too. Happy to take this.

PR #496, opened 2026-08-04
site: fix redirects and broken links

XhmikosR:
Only the `sites.google.com` and `docs.webpagetest.org` are left. Not sure how to proceed with those... There are some other weird ones like `g.co/chromeuxdash`.

### DIFF
--- a/site/src/_data/site.js
+++ b/site/src/_data/site.js
@@ -35,12 +35,12 @@ module.exports = () => {
       {
         title: 'magento',
         logoFileName: 'magento.svg',
-        url: 'https://marketplace.magento.com/rafaelcg-magento2-quicklink.html',
+        url: 'https://github.com/rangerz/magento2-module-quicklink',
       },
       {
         title: 'react',
         logoFileName: 'react.svg',
-        url: 'https://github.com/HOUCe/react-quicklink-component/',
+        url: 'https://web.dev/articles/quicklink',
       },
       {
         title: 'angular',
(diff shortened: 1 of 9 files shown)


site/src/_data/site.js
```diff
@@ -35,12 +35,12 @@ module.exports = () => {
       {
         title: 'magento',
         logoFileName: 'magento.svg',
-        url: 'https://marketplace.magento.com/rafaelcg-magento2-quicklink.html',
+        url: 'https://github.com/rangerz/magento2-module-quicklink',
       },
       {
         title: 'react',
         logoFileName: 'react.svg',
-        url: 'https://github.com/HOUCe/react-quicklink-component/',
```
XhmikosR, line 43: this repo is deleted, unsure if the `web.dev` article is suitable. Let me know and I can drop it altogether.

site/src/_data/site.js
```diff
@@ -50,7 +50,7 @@ module.exports = () => {
       {
         title: 'vue',
         logoFileName: 'vue.svg',
-        url: 'https://nuxtjs.org/api/components-nuxt-link/',
+        url: 'https://nuxt.com/docs/api/components/nuxt-link',
```
XhmikosR, line 53: Went with this that always points to the latest version via a meta refresh

addyosmani (approved):
Thank you for the link sweep — these were overdue. The trailing-slash fixes on the internal links (`/api` → `/api/` and friends) are a nice touch; they save a redirect hop on every click, which matters more now that the site's behind a CDN. On the `sites.google.com` / `docs.webpagetest.org` stragglers you flagged — don't hold this PR for them, let's drop or replace those separately. Merging.

PR #501, opened 2026-09-10
site: use unique page descriptions

### LINKED ISSUE #366: site: use unique page descriptions

Tiancheng-Xu:
Fixes #366. The API, Approach, Demo, and Measure pages now provide page-specific meta descriptions through their existing front matter metadata path. Added a build-output regression check to prevent those pages from reusing the same generic description. Tests: site npm test (stylelint, production Eleventy build, description check); root npm test (build-all, uvu 19 passed, 1 skipped); npm run lint; npm run size (all limits passed); git diff --check.


XhmikosR:
I don't think we need the test file, just the descriptions.

Tiancheng-Xu:
Addressed the review: removed the added description-check script and package test hook, keeping the PR focused on the four page descriptions only. `git diff --check` passes; the local dependency install/test was not run because node_modules is absent in the isolated checkout.

